PR #50, opened 2020-09-17
WIP: Feature/multiproofs

askolesov:


utils/range_helper/range.go
```diff
@@ -0,0 +1,93 @@
...
+
+type RangeResult struct {
+	Total int               `json:"total"`
+	Items []json.RawMessage `json:"items"`
+}
+
+func QueryRangeWithTotal(ctx Context, storeKey string, prefix []byte, params pagination.RangeParams,
+	totalKey []byte, valueUnmarshaler func([]byte) json.RawMessage) (RangeResult, int64, error) {
+	// Is all range queried
+	isAllRange := false
```
ashcherbakov, line 36: Do we ever set it to true?

utils/cli/cli.go
```diff
@@ -92,41 +95,68 @@ func (ctx CliContext) WithHeight(height int64) CliContext {
...
-	ctx.context = ctx.context.WithHeight(0)
 
 	return ctx.context.QueryStore(key, storeName)
 }
 
+func (ctx CliContext) QueryStoreAtHeight(height int64, key []byte, storeName string) ([]byte, int64, error) {
+	return ctx.context.WithHeight(height).QueryStore(key, storeName)
+}
+
+func (ctx CliContext) QueryRange(startKey []byte, endKey []byte, limit int,
```
ashcherbakov, line 144: Do we call this method somewhere? It looks like we always call `QueryRangeWithTotalAndHandleIO` only, don't we?

docs/transactions.md
```diff
@@ -1377,12 +1377,28 @@ The account is not revoked until sufficient number of Trustees approve it.
 
 Gets all proposed but not approved accounts.
```
ashcherbakov, line 1379: Please add the information about the new keys (account counters) to the `KV Store` section.

integration_tests/cli/auth-demo.sh
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
```
ashcherbakov, line 15: Do we have tests for negative/malicious cases:
1) The returned number of items (all accounts) isn't equal to the total number (of accounts)?
2) Invalid element is returned as part of the range request?
3) The returned item(s) are valid, but not within the range.
4) Not all elements from the range are returned.

askolesov:
Closing due inactivity

PR #53, opened 2020-09-22
Added Version and OTA-related fields to ModelInfo

spivachuk:
- Added Version field and OTA-related fields to ModelInfo.
- Rearranged fields in ModelInfo and related types in a single order.
- Added more tests for ModelInfo.
- Added more checks to existing tests of ModelInfo.
- Made corrections in the existing code.


askolesov (approved):

x/modelinfo/handler.go
```diff
@@ -86,35 +90,29 @@ func handleMsgUpdateModelInfo(ctx sdk.Context, keeper keeper.Keeper, authKeeper
 		return err.Result()
 	}
 
+	if msg.OtaURL != "" && modelInfo.OtaURL == "" {
+		return types.ErrOtaURLCannotBeSet(msg.VID, msg.PID).Result()
```
ashcherbakov, line 94: Do we have unit and integration tests for the case of updating the OTA URL?

PR #55, opened 2020-09-29
Added instruction for genesis node deployment

askolesov:


docs/running-genesis-node.md
```diff
@@ -0,0 +1,161 @@
...
+    * Remember generated `address` and `pubkey` they will be used later. 
+    You can retrieve `address` and `pubkey` values anytime using `dclcli keys show <name>`. 
+    Of course, only on the machine where the keypair was generated.
+
+4. Prepare genesis node configuration:
+
+    * Initialize new configuration: `dcld init <node-name> --chain-id dclchain`.
+    * Add genesis account with the generated key and `Trustee`, `NodeAdmin` roles:
+    `dcld add-genesis-account --address=<address> --pubkey=<pubkey> --roles="Trustee,NodeAdmin"`
+    * Optionally, add other genesis accounts using the same command.
```
ashcherbakov, line 64: I think we need to be more explicit that the steps above must be done by every genesis node. 
So, we can say here, that we need to wait until all genesis nodes do the steps above and share their accounts with each other.

docs/running-genesis-node.md
```diff
@@ -0,0 +1,161 @@
...
+        }
+        ```
+    * Remember generated `address` and `pubkey` they will be used later. 
+    You can retrieve `address` and `pubkey` values anytime using `dclcli keys show <name>`. 
+    Of course, only on the machine where the keypair was generated.
+
+4. Prepare genesis node configuration:
+
+    * Initialize new configuration: `dcld init <node-name> --chain-id dclchain`.
+    * Add genesis account with the generated key and `Trustee`, `NodeAdmin` roles:
```
ashcherbakov, line 62: For the Test Net it can be fine, but for other Nets we may not want to mix NodeAdmin and Trustee accounts. Probably we should mention this explicitly.

docs/running-genesis-node.md
```diff
@@ -0,0 +1,161 @@
...
+    `dcld add-genesis-account --address=<address> --pubkey=<pubkey> --roles="Trustee,NodeAdmin"`
+    * Optionally, add other genesis accounts using the same command.
+    * Create genesis transaction: `dcld gentx --from <name>`
```
ashcherbakov, line 65: <name> here is the name our NodeAdmin account? I think we need to clarify this.

docs/running-genesis-node.md
```diff
@@ -0,0 +1,161 @@
...
+    * In order to ensure that account is created and has assigned role you can use the command: 
+    `dclcli query auth account --address=<address>`.
+    Expected output format: 
+        ```json
+        {
+          "result": {
+            "address": string, // bench32 encoded address
+            "public_key": "string, // bench32 encoded public key
+            "roles": [
+              "NodeAdmin"
```
ashcherbakov, line 102: I think we specified a Trustee role for our account as well

docs/running-genesis-node.md
```diff
@@ -0,0 +1,161 @@
...
+            "pub_key": {
+              "type": string,
+              "value": string
+            },
+            "voting_power": string
+          }
+        }
+        ```
+    
+    * Get the list of nodes participating in the consensus for the last block: `dclcli tendermint-validator-set`.
```
ashcherbakov, line 158: Should we do this step only after all the genesis nodes are started up (finish the steps above)?

docs/running-genesis-node.md
```diff
@@ -0,0 +1,161 @@
...
+    Of course, only on the machine where the keypair was generated.
+
+4. Prepare genesis node configuration:
+
+    * Initialize new configuration: `dcld init <node-name> --chain-id dclchain`.
+    * Add genesis account with the generated key and `Trustee`, `NodeAdmin` roles:
+    `dcld add-genesis-account --address=<address> --pubkey=<pubkey> --roles="Trustee,NodeAdmin"`
+    * Optionally, add other genesis accounts using the same command.
+    * Create genesis transaction: `dcld gentx --from <name>`
+    * Collect genesis transactions: `dcld collect-gentxs`.
```
ashcherbakov, line 66: How we are going to get Validator genesis transactions from other nodes? Does `collect-gentxs` do it?

docs/running-genesis-node.md
```diff
@@ -0,0 +1,161 @@
...
+    * Validate genesis file: `dcld validate-genesis`.
+    * Genesis file is located in `$HOME/.dcld/config/genesis.json`. Give this file to each new node admin.
```
ashcherbakov, line 68: By `new admin` we mean non-genesis node admins, don't we? We may point to the `running-node.md` doc here.

ashcherbakov (approved):

PR #57, opened 2020-09-30
Extend instructions for deployment of a new network

ashcherbakov:
- Extends docs related to deployment
- Created `deployment` folder. It contains files and templates to be used for deployment, as well as subfolders with configurations (genesis and persistent_peers) for every persistent chain (such as Test Net). The subfolder name is equal to the chain-id.
- Moved `ansible` folder with scripts to deploy a custom net to the `deployment` folder


docs/cli-help.md
```diff
@@ -52,7 +52,7 @@ Roles:
 - `Vendor` - Is able to add models.
 - `TestHouse` - Is able to add testing results for a model.
 - `ZBCertificationCenter` - Is able to certify and revoke models.
-- `NodeAdmin` - Is able to add nodes to validator pool.
+- `NodeAdmin` - Is able to add nodes to the network of validator nodes.
```
askolesov, line 55: Is able to add validator nodes to the network.

probably sounds better

PR #70, opened 2020-10-29
Fixed cetrify model

adenishchenko:


x/compliance/handler.go
```diff
@@ -155,3 +160,23 @@ func checkZbCertificationRights(ctx sdk.Context, authKeeper auth.Keeper, signer
...
+func checkZbCertificationDone(
+	ctx sdk.Context,
+	keeper keeper.Keeper,
+	authKeeper auth.Keeper,
+	signer sdk.AccAddress,
+	msg types.MsgCertifyModel) sdk.Error {
+	if keeper.IsComplianceInfoPresent(ctx, msg.CertificationType, msg.VID, msg.PID) {
+		complianceInfo := keeper.GetComplianceInfo(ctx, msg.CertificationType, msg.VID, msg.PID)
+		if complianceInfo.State == types.Certified {
+			if bytes.Equal(complianceInfo.Owner, signer) {
```
ashcherbakov, line 173: Just a small comment about code readability:
it's usually a best practice to avoid more than two nested if-else conditions (personally, I prefer not to have more than 1).
So, the code can be re-written as follows:
```
if !keeper.IsComplianceInfoPresent(ctx, msg.CertificationType, msg.VID, msg.PID) {
        return nil
}
complianceInfo := keeper.GetComplianceInfo(ctx, msg.CertificationType, msg.VID, msg.PID)
if complianceInfo.State != types.Certified {
        retrun nil
}
if bytes.Equal(complianceInfo.Owner, signer) {
        return nil
}
return types.ErrAlreadyCertifyed(msg.VID, msg.PID)
```

ashcherbakov (approved):

ashcherbakov (approved):

PR #87, opened 2021-06-23
Create a separate model type

ankur325:
Refactor the code to make it easier to update the model type in future.


ankur325:
Adding the commits from this branch into the earlier PR.

PR #107, opened 2021-11-15
Adding more validation test cases

ankur325:
Added few more test cases and streamlined validation for both Model and ModelVersion. Also added few new validations based on recent spec changes e.g. only https URL is allowed when HTTP protocol is used.

### DIFF
--- a/x/model/internal/types/msgs_test.go
+++ b/x/model/internal/types/msgs_test.go
@@ -16,6 +16,7 @@
 package types
 
 import (
+	"strings"
 	"testing"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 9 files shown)


x/model/internal/types/msgs_test.go
```diff
@@ -159,3 +160,254 @@ func TestMsgUpdateModelGetSignBytes(t *testing.T) {
...
+	// otaFileSize is mandatory if otaURL is set
+	modelVersion = getTestModelVersion()
+	modelVersion.OtaURL = testconstants.ValidHTTPSURL
+	modelVersion.OtaFileSize = 0
+	require.NotNil(t, NewMsgAddModelVersion(modelVersion, testconstants.Signer).ValidateBasic())
+
+	// otaFileSize is optional if otaURL is not set
+	modelVersion = getTestModelVersion()
+	modelVersion.OtaURL = ""
+	modelVersion.OtaFileSize = 0
```
ashcherbakov, line 242: Looks like execution (require.Nil...) is missed here

ashcherbakov (approved):

PR #119, opened 2021-12-06
adds better automation for deployment routine

andkononykhin:


deployment/scripts/verify_node
```diff
@@ -0,0 +1,43 @@
...
+
+export PATH="$PATH:/usr/bin"
+
+echo "Checking the node 'sync_info.latest_block_height' value is not increasing..."
+
+echo "sync_info.latest_block_height Try 1"
+check1="$(dclcli status --node "$NODE_IP" | grep latest_block_height)"
+
+echo "sync_info.latest_block_height Try 2"
+sleep 6
```
ashcherbakov, line 31: We may use a different value for empty block creation (currently it's 5 secs), so that check may need to be changed.

docs/running-node.md
```diff
@@ -0,0 +1,144 @@
+# Running a DCLedger Node
+
+This document describes in details how to configure different types of DCLedger nodes.
+
+## Components
+
+*   Common release artifacts:
+    * Binary artifacts (part of the release):
+        * dcld: The binary used for running a node.
+        * dclcli: The binary that allow users to interact with the network of nodes.
```
ashcherbakov, line 10:
```suggestion
        * dclcli: The binary that allows users to interact with the network of nodes.
```

docs/running-node.md
```diff
@@ -0,0 +1,144 @@
...
+- x64 2.0 GHz 2v CPU
+
+## Operating System
+
+Current delivery is compiled and tested under `Ubuntu 18.04.3 LTS` so we recommend using this distribution for now. In future, it will be possible to compile the application for a wide range of operating systems thanks to Go language.
+
+Notes:
+    * the deployment commands below will try to setup and run `dcld` systemd service on Ubuntu
+    * that will require `sudo` for a user
+    * in case non-Ubuntu system these steps will be scipped so you would need to take case about
```
ashcherbakov, line 41:
```suggestion
    * in case non-Ubuntu system these steps will be scipped so you would need to take care about
```

docs/running-node.md
```diff
@@ -0,0 +1,144 @@
...
+        - by default it will evaluate it from the `persisten_peers.txt` (using its first entry)
+        - you can specify the peer explicitly using `-p/--peer`: `-p <node-id@host:port>` (port is usually considered as `26656` here)
+    - generates a new key entry for a node admin account
+        - by default `<node_name>admin` key name is used
+        - can be configured using `-k/--key-name` option
+    - properly locates `genesis.json`
+    - configures and starts the node
+
+Outputs:
+    - key data (address, public key, mnemonic) as `*.dclkey.json` file in the current directory
```
ashcherbakov, line 100: Is it also Node Admin key data?

docs/running-node.md
```diff
@@ -0,0 +1,144 @@
...
+   an account with `NodeAdmin` role. And **wait** until:
+
+   * Account is created
+   * The node completed a catch-up:
+        * `dclcli status --node <node ip>` returns `false` for `catching_up` field
+
+3. Make the node a validator:
+
+```bash
+$ dclcli tx validator add-node \
```
ashcherbakov, line 114: What about a passphrase used for account creation? Will the scripts prompt for it?

deployment/scripts/run_dcl_node
```diff
@@ -0,0 +1,459 @@
...
+  echo "\
+Usage: $0 [options] node-name
+
+Initializes DCL node of a given type.
+
+Options:
+  -c,  --chain-id   chain id, default: '$DEF_CHAIN_ID'
+  -h,  --help       print this help and exit
+  -k,  --key-name   account key-name, default: '<node-name>admin'
+  -p,  --peer       peer hostspec as 'HOST:PORT', default: first entry of 'persistent_peers.txt'
```
ashcherbakov, line 74: Not sure that this is a correct default. I believe the local CLI should be always connected to localhost, shouldn't it?
If it's not a localhost, then it should be the node owned by that user/company, and not the first entry from `persistent_peers.txt`.

deployment/scripts/run_dcl_node
```diff
@@ -0,0 +1,459 @@
...
+    #parsed_args+="\n\trelease=$pkg_version"
+
+    #echo -e "Parsed arguments:\n$parsed_args"
+fi
+
+echo "Configuring CLI"
+
+if [[ "$NODE_TYPE" == "observer" ]]; then
+    # FIXME is it right
+    config_cli "$CHAIN_ID"
```
ashcherbakov, line 386: We need a working local CLI in case of Observer as well, don't we? For example, to check the node status.
So, we need to specify peer there (or use localhost).

deployment/scripts/run_dcl_node
```diff
@@ -0,0 +1,459 @@
...
+        echo -e "\tUse 'systemctl status dcld' to get the node service status."
+        echo "Use 'journalctl -u dcld.service -f' to see node logs."
+    else
+        dcld start &
+        echo "Node has been stared in the backgroud."
+    fi
+}
+
+
+function verify_node {
```
ashcherbakov, line 304: Are these functions used?

deployment/scripts/run_dcl_node
```diff
@@ -0,0 +1,459 @@
...
+        # config the service
+        # Ubuntu only
+        grep -q -i -e 'ubuntu' /etc/*-release && {
+            if [[ ! -f "$_dcld_service_p" ]]; then
+                echo "Error. '$_dcld_service_p' file not found"
+                exit 1
+            fi
+
+            # Open '26656' (p2p) and '26657' (RPC) ports. 
+            sudo ufw allow 26656/tcp
```
ashcherbakov, line 268: Makes sense to move to a separate script

deployment/scripts/verify_node
andkononykhin: added TODO comment there

docs/running-node.md
andkononykhin: fixed

docs/running-node.md
andkononykhin: fixed

docs/running-node.md
andkononykhin: yes, updated

docs/running-node.md
andkononykhin: yes

deployment/scripts/run_dcl_node
andkononykhin: hid that for now

deployment/scripts/run_dcl_node
andkononykhin: kept that as default setting, cli should communicate with  the local node

deployment/scripts/run_dcl_node
andkononykhin: no for now, there is a separate script that may help to do some checks but out of the current script logic

deployment/scripts/run_dcl_node
andkononykhin: added TODO there

ashcherbakov (approved):

PR #127, opened 2021-12-09
[WIP][issue 99] integration cleanup

andkononykhin:


cmd/dcld/cmd/genaccounts.go
```diff
@@ -40,43 +43,40 @@ the address will be looked up in the local Keybase.
...
 			serverCtx := server.GetServerContextFromCmd(cmd)
 			config := serverCtx.Config
 
 			config.SetRoot(clientCtx.HomeDir)
 
 			addr, err := sdk.AccAddressFromBech32(viper.GetString(FlagAddress))
-			if err != nil {
-				return err
-			}
+			//if err != nil {
```
ashcherbakov, line 55: Is it intended?

x/dclauth/keeper/account.go
```diff
@@ -82,7 +113,7 @@ func (k Keeper) IterateAccounts(ctx sdk.Context, cb func(account types.Account)
 
 // Check if account has assigned role.
 func (k Keeper) HasRole(ctx sdk.Context, addr sdk.AccAddress, roleToCheck types.AccountRole) bool {
-	account, found := k.GetAccount(ctx, addr)
+	account, found := k.GetAccountO(ctx, addr)
```
ashcherbakov, line 116: Why can't we use `GetAccount` instead so that returned `nil` is the same as `found=false`?

ashcherbakov (approved):

PR #137, opened 2021-12-21
Migrating PKI module

ashcherbakov:
- All PKI commands are migrated
- CLI integration tests pass
- Cert chain command is not supported intentionally (there were some concerns in community and it's not needed for Matter)

TBD (next PRs):
- gRPC/REST tests
- Unit tests
- Improve get-all queries to not return the whole pem cert and use store-based queries for the future light client


proto/pki/proposed_certificate_revocation.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+package zigbeealliance.distributedcomplianceledger.pki;
+
+option go_package = "github.com/zigbee-alliance/distributed-compliance-ledger/x/pki/types";
+
+import "cosmos_proto/cosmos.proto";
+
+message ProposedCertificateRevocation {
+  string subject = 1;
```
andkononykhin, line 9: as an option you may consider here to make an embedded composition with `CerificateIdentifier` like [here](https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/4b152049bf86e7d6dfc83183664a90826177dc9f/proto/dclauth/account.proto#L16).

Generally it wouldn't affect code with getters, for sure code with constructors would be impacted. [Example](https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/4b152049bf86e7d6dfc83183664a90826177dc9f/x/dclauth/keeper/msg_server_propose_add_account.go#L67-L68).

config.yml
```diff
@@ -0,0 +1,16 @@
+accounts:
+  - name: alice
```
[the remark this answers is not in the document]
ashcherbakov: Without that file scaffolding outputs errors for me (although looks like generated necessary files). 
Added a TODO comment. Needs to be researched.

x/pki/client/cli/query_approved_certificates.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/spf13/viper"
+	"github.com/zigbee-alliance/distributed-compliance-ledger/x/pki/types"
+)
+
+func CmdListApprovedCertificates() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "all-x509-certs",
+		Short: "Gets all certificates (root, intermediate and leaf)",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
```
andkononykhin, line 18: does the command respect configuration (e.g. output=json or output=text) in client.toml?

x/pki/module.go
```diff
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 
-	// this line is used by starport scaffolding # 1
+	"context"
```
andkononykhin: could you please rename `capability` word (module name) with `pki`? (looks like `starport` has a minor issue and uses the name inside a template)

config.yml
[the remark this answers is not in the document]
andkononykhin: Thanks

andkononykhin (commented):
Generally looks good to me. Please find my  comments though.

x/pki/module.go
ashcherbakov: Renamed, thanks.

PR #138, opened 2021-12-22
Improved queries response when key is absent

ashcherbakov:
- return empty list or null instead of an error
 - made query proto responses nullable


x/pki/client/cli/query_child_certificates.go
```diff
@@ -6,58 +6,26 @@ import (
...
-
-	return cmd
-}
-
 func CmdShowChildCertificates() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "show-child-certificates [issuer] [authority-key-id]",
-		Short: "shows a ChildCertificates",
-		Args:  cobra.ExactArgs(2),
+		Use:   "all-child-x509-certs",
```
andkononykhin, line 15: as  understand we don't provide a cli to get all child certificates for any given certificates. It's ok but the current single-item command naming contradicts with other modules (auth, validator) since `all-...` is used there for list queries only.

andkononykhin (commented):
The change looks good. Only one concern regarding the naming. Please check.

x/pki/client/cli/query_child_certificates.go
andkononykhin: it has been discussed, no issue here

x/pki/client/cli/utils.go
```diff
@@ -17,3 +20,17 @@ func ReadFromFile(target string) (string, error) {
 		return target, nil
 	}
 }
+
+func HandleError(err error) error {
```
andkononykhin, line 24: shouldn't we expose that on common level so all modules will re-use?

andkononykhin (approved):

PR #156, opened 2021-12-30
Migrated PKI Handler unit tests

spivachuk:


x/pki/handler_test.go
```diff
@@ -0,0 +1,1522 @@
...
+	// approve
+	approveAddX509RootCert := types.NewMsgApproveAddX509RootCert(
+		setup.Trustee.String(), testconstants.RootSubject, testconstants.RootSubjectKeyID)
+	_, err = setup.Handler(setup.Ctx, approveAddX509RootCert)
+	require.NoError(t, err)
+
+	// approve second time
+	_, err = setup.Handler(setup.Ctx, approveAddX509RootCert)
+	require.Error(t, err)
+	require.True(t, sdkerrors.ErrUnauthorized.Is(err))
```
ashcherbakov, line 394: As I can see, it was implemented to raise ErrUnauthorized in legacy as well, but it looks weird. Probably better to create and raise a special error code in `pki/types/errors.go`, something like `ErrProposedCertificateAlreadyApproved`.

x/pki/handler_test.go
```diff
@@ -0,0 +1,1522 @@
...
+		setup.Trustee.String(), testconstants.RootSubject, testconstants.RootSubjectKeyID)
+	_, err := setup.Handler(setup.Ctx, proposeRevokeX509RootCert)
+	require.NoError(t, err)
+
+	// approve revocation by the same trustee
+	approveRevokeX509RootCert := types.NewMsgApproveRevokeX509RootCert(
+		setup.Trustee.String(), testconstants.RootSubject, testconstants.RootSubjectKeyID)
+	_, err = setup.Handler(setup.Ctx, approveRevokeX509RootCert)
+	require.Error(t, err)
+	require.True(t, sdkerrors.ErrUnauthorized.Is(err))
```
ashcherbakov, line 1066: The same as in previous comment.

ashcherbakov (approved):

PR #166, opened 2022-01-12
Disable Gas

ashcherbakov:
- disabled Gas in DCL
- add dev docs with a short how-to add a new module
- remove dry-run flag from the CLI
- add helper scripts to run integration tests


README-DEV.md
```diff
@@ -144,6 +144,49 @@ Please take into account the following when sending a PR:
...
+         return err
+     }
+    ```
+- Implement business logic in `msg_server_xxx.go`
+- Add unit tests (see other modules for reference)
+- Add CLI-based integration tests to `integration_tests/cli/<module>` (see other modules for reference)
+- Add gRPC/REST-based integration tests to `integration_tests/grpc_rest/<module>` (see other modules for reference)
+
+## How To Make Changes in Data Model for Existing Modules
+- Do the changes in `.proto` files. Never change `.pb` file manually.
```
andkononykhin, line 183: I think makes sense to emphasize that more like: **Never change `.pb` file manually.**

integration_tests/start-pool.sh
```diff
@@ -0,0 +1,24 @@
...
+
+  for NODE_CONFIG in ${NODE_CONFIGS}; do
+    sed -i $SED_EXT 's/timeout_propose = "3s"/timeout_propose = "500ms"/g' "${NODE_CONFIG}"
+    sed -i $SED_EXT 's/timeout_prevote = "1s"/timeout_prevote = "500ms"/g' "${NODE_CONFIG}"
+    sed -i $SED_EXT 's/timeout_precommit = "1s"/timeout_precommit = "500ms"/g' "${NODE_CONFIG}"
+    sed -i $SED_EXT 's/timeout_commit = "5s"/timeout_commit = "500ms"/g' "${NODE_CONFIG}"
+  done
+}
+
+make localnet_clean
```
andkononykhin, line 17: as an option these 4 lines can be replaced with:
```bash
make install image localnet_rebuild
```

integration_tests/start-pool.sh
```diff
@@ -0,0 +1,24 @@
+source integration_tests/cli/common.sh
+
+LOCALNET_DIR=".localnet"
+SED_EXT=
+
+patch_consensus_config() {
```
andkononykhin, line 6: I think makes sense to move that func to common.sh so both run-all and start-pool may use it

x/dclauth/types/constants.go
```diff
@@ -16,9 +16,9 @@ package types
 
 // Default parameter values.
 const (
-	MaxMemoCharacters             uint64  = 256
-	TxSizeCostPerByte             uint64  = 10
-	DefaultSigVerifyCostED25519   uint64  = 590
-	DefaultSigVerifyCostSecp256k1 uint64  = 1000
-	AccountApprovalPercent        float64 = 0.66
+	DclMaxMemoCharacters      uint64  = 256
```
andkononykhin, line 19: shouldn't we refer to [default cosmos'](https://github.com/cosmos/cosmos-sdk/blob/2f335a2b8cc4fcfa337c104823fab9f6ecb387da/x/auth/types/params.go#L13) value here (as for `DefaultTxSigLimit`)

x/dclauth/keeper/account.go
```diff
@@ -166,8 +166,12 @@ func (k Keeper) GetModuleAddress(moduleName string) sdk.AccAddress {
 	return nil
 }
 
-// just a stub to have AccountKeeper.GetParams API filled.
 func (k Keeper) GetParams(ctx sdk.Context) (params authtypes.Params) {
-	params = authtypes.DefaultParams()
-	return
+	return authtypes.Params{
+		MaxMemoCharacters:      types.DclMaxMemoCharacters,
+		TxSigLimit:             authtypes.DefaultTxSigLimit,
```
andkononykhin, line 172: I suggest to use our own constant which refers to that cosmos' value

andkononykhin (commented):
looks good, just a few suggestions please

andkononykhin (approved):

PR #177, opened 2022-01-13
Compliancetest unit tests

DenisRybas:


x/compliancetest/types/message_add_testing_result_test.go
```diff
@@ -1,42 +1,135 @@
...
+			name: "pid > 65535",
+			msg: MsgAddTestingResult{
+				Signer:                sample.AccAddress(),
+				Pid:                   65536,
+				Vid:                   1,
+				SoftwareVersionString: testconstants.SoftwareVersionString,
+				TestDate:              testconstants.CertificationDate.Format(time.RFC3339),
+				TestResult:            testconstants.TestResult,
+			},
+			err: validator.ErrFieldUpperBoundViolated,
```
ashcherbakov, line 100: Please also add the following negative tests:
1) TestDate is not set (empty string). Expected: ErrRequiredFieldMissing
2) TestResult is not set (empty string). Expected: ErrRequiredFieldMissing
3) TestDate doesn't have expected format RFC3339)

x/compliancetest/types/message_add_testing_result_test.go
```diff
@@ -1,42 +1,135 @@
...
+		},
+	}
+
+	positive_tests := []struct {
+		name string
+		msg  MsgAddTestingResult
+	}{
+		{
 			name: "valid address",
```
ashcherbakov, line 109: Please find a better description, as not only address, but all fields are valid here.

x/compliancetest/types/message_add_testing_result_test.go
```diff
@@ -1,42 +1,135 @@
...
 		name string
 		msg  MsgAddTestingResult
 		err  error
 	}{
 		{
 			name: "invalid address",
 			msg: MsgAddTestingResult{
-				Signer: "invalid_address",
+				Signer:                "invalid_address",
+				SoftwareVersionString: testconstants.SoftwareVersionString,
```
ashcherbakov, line 24: It's better to set `SoftwareVersion` everywhere explicitly, and add one explicit positive test where SoftwareVersion=0.

x/compliancetest/types/message_add_testing_result_test.go
```diff
@@ -1,42 +1,135 @@
...
+				TestDate:              testconstants.CertificationDate.Format(time.RFC3339),
+				TestResult:            testconstants.TestResult,
 			},
 			err: sdkerrors.ErrInvalidAddress,
-		}, {
+		},
+		{
+			name: "vid is 0",
```
ashcherbakov, line 31: Please remove `msgs_test.go` as these tests duplicate them (and these tests are better than ones from `msgs_test.go`).
Please make sure that all cases from `msgs_test.go` are implemented here.

PR #199, opened 2022-01-27
Validations update

DenisRybas:


proto/compliance/tx.proto
```diff
@@ -21,10 +21,10 @@ message MsgCertifyModel {
   int32 vid = 2 [(gogoproto.moretags) = "validate:\"gte=1,lte=65535\""];
   int32 pid = 3 [(gogoproto.moretags) = "validate:\"gte=1,lte=65535\""];
   uint32 software_version = 4;
-  string software_version_string = 5 [(gogoproto.moretags) = "validate:\"required\""];
+  string software_version_string = 5 [(gogoproto.moretags) = "validate:\"required,max=10\""];
```
ashcherbakov, line 24: I believe `software_version_string` should be 64 as in ModelVersion and the spec (please fix for all occurrences)

proto/compliance/tx.proto
```diff
@@ -21,10 +21,10 @@ message MsgCertifyModel {
...
   uint32 c_d_version_number = 6  [(gogoproto.moretags) = "validate:\"gte=0,lte=65535\""];
   string certification_date = 7 [(gogoproto.moretags) = "validate:\"required\""];
-  string certification_type = 8 [(gogoproto.moretags) = "validate:\"required\""];
+  string certification_type = 8 [(gogoproto.moretags) = "validate:\"required,max=100\""];
   string reason = 9;
```
ashcherbakov, line 28: Please add max restriction for the `reason` field.

proto/compliancetest/tx.proto
```diff
@@ -19,8 +19,8 @@ message MsgAddTestingResult {
   int32 vid = 2 [(gogoproto.moretags) = "validate:\"gte=1,lte=65535\""];
   int32 pid = 3 [(gogoproto.moretags) = "validate:\"gte=1,lte=65535\""];
   uint32 software_version = 4;
-  string software_version_string = 5 [(gogoproto.moretags) = "validate:\"required\""];
-  string test_result = 6 [(gogoproto.moretags) = "validate:\"required\""];
+  string software_version_string = 5 [(gogoproto.moretags) = "validate:\"required,max=10\""];
+  string test_result = 6 [(gogoproto.moretags) = "validate:\"required,max=256\""];
```
ashcherbakov, line 23: I believe test result may contain quite a lot of info, so 256 is too small.
Let's constrain it to 10 MB, so max=`10*1024*1024`

proto/pki/tx.proto
```diff
@@ -21,51 +21,51 @@ service Msg {
 
 message MsgProposeAddX509RootCert {
   string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.moretags) = "validate:\"required\""];
-  string cert = 2 [(gogoproto.moretags) = "validate:\"required\""];
+  string cert = 2 [(gogoproto.moretags) = "validate:\"required,max=3500000\""];
```
ashcherbakov, line 24: Let's constraint to 10MB (`10 * 1024 * 1024`)

proto/pki/tx.proto
```diff
@@ -21,51 +21,51 @@ service Msg {
...
 
 message MsgProposeAddX509RootCertResponse {
 }
 
 message MsgApproveAddX509RootCert {
   string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.moretags) = "validate:\"required\""];
-  string subject = 2 [(gogoproto.moretags) = "validate:\"required\""];
-  string subject_key_id = 3 [(gogoproto.moretags) = "validate:\"required\""];
+  string subject = 2 [(gogoproto.moretags) = "validate:\"required,max=50\""];
+  string subject_key_id = 3 [(gogoproto.moretags) = "validate:\"required,max=150\""];
```
ashcherbakov, line 33: According to https://www.rfc-editor.org/rfc/rfc3280#section-4.2.1.2
subject key ID should be either 160 bit or 60 bit. So, 150 may be too small.
It looks like other ways of generation of subject_key_id are also possible.
So, let's assign `max=256`.

proto/pki/tx.proto
```diff
@@ -21,51 +21,51 @@ service Msg {
...
 }
 
 message MsgProposeAddX509RootCertResponse {
 }
 
 message MsgApproveAddX509RootCert {
   string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.moretags) = "validate:\"required\""];
-  string subject = 2 [(gogoproto.moretags) = "validate:\"required\""];
-  string subject_key_id = 3 [(gogoproto.moretags) = "validate:\"required\""];
+  string subject = 2 [(gogoproto.moretags) = "validate:\"required,max=50\""];
```
ashcherbakov, line 32: Too small. Look at https://crypto.stackexchange.com/questions/54148/does-maximum-length-of-cn-in-a-x509-cert-include-the-length-of-the-oids and https://www.rfc-editor.org/rfc/rfc3280#section-4.1.2.6.
Let's restrict to some reasonably high value. For example, `1024` (1KB)

ashcherbakov (approved):

PR #209, opened 2022-02-01
Add LsfURL and LsfRevision Fields

ankur325:
Added the LsfURL and LsfRevision fields to the Device Model


x/model/keeper/msg_server_model.go
```diff
@@ -125,6 +127,15 @@ func (k msgServer) UpdateModel(goCtx context.Context, msg *types.MsgUpdateModel)
 		model.ProductUrl = msg.ProductUrl
 	}
 
+	if msg.LsfUrl != "" && msg.LsfUrl != model.LsfUrl {
```
ashcherbakov, line 130: What if `msg.LsfUrl == model.LsfUrl` and `msg.LsfRevision != model.LsfRevision`? It looks like now it won't be updated.
Should we either return an error in this case, or apply the new `LsfRevision` for the same `LsfUrl`?


Another case: only `LsfRevision` is set (so that msg.LsfUrl == "").

x/model/types/messages_model_test.go
```diff
@@ -283,6 +283,38 @@ func TestMsgCreateModel_ValidateBasic(t *testing.T) {
...
+			}(validMsgCreateModel()),
+			err: validator.ErrFieldMaxLengthExceeded,
+		},
+		{
+			name: "LsfRevision is missing",
+			msg: func(msg *MsgCreateModel) *MsgCreateModel {
+				msg.LsfRevision = 0
+				return msg
+			}(validMsgCreateModel()),
+			err: validator.ErrRequiredFieldMissing,
```
ashcherbakov, line 316: It would be great to also add the following (negative) tests:
- LsfRevision < 0
- LsfRevision > 65535

x/model/types/messages_model_test.go
```diff
@@ -498,6 +530,20 @@ func TestMsgCreateModel_ValidateBasic(t *testing.T) {
 				return msg
 			}(validMsgCreateModel()),
 		},
+		{
+			name: "LsfUrl is omitted",
```
ashcherbakov, line 534: I think we need to add the same cases (negative and positive) for `LsfUrl` and `LsfRevision` for `UpdateModel`

ashcherbakov (commented):
Looks good, please have a look at couple of comments.

proto/model/tx.proto
```diff
@@ -55,6 +57,8 @@ message MsgUpdateModel {
   string user_manual_url = 10 [(gogoproto.moretags) = "validate:\"omitempty,url,startsnotwith=http:,max=256\""];
   string support_url = 11 [(gogoproto.moretags) = "validate:\"omitempty,url,startsnotwith=http:,max=256\""];
   string product_url = 12 [(gogoproto.moretags) = "validate:\"omitempty,url,startsnotwith=http:,max=256\""];
+  string lsf_url = 13 [(gogoproto.moretags) = "validate:\"required_unless=LsfRevision 0,omitempty,url,startsnotwith=http:,max=256\""];
+  int32 lsf_revision = 14 [(gogoproto.moretags) = "validate:\"required_with=LsfUrl,gte=0,lte=65535\""];
```
ashcherbakov, line 61: According to https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/service_device_management/DistributedComplianceLedger.adoc#1512-lsfrevision,
> this value SHALL monotonically increase by 1. 

Should we add this check to DCL (that a new_value=old_value+1 for the same lsfUrl)? BTW `handler_test` can be used for a unit test here.

proto/model/tx.proto
```diff
@@ -38,6 +38,8 @@ message MsgCreateModel {
   string user_manual_url = 14 [(gogoproto.moretags) = "validate:\"omitempty,url,startsnotwith=http:,max=256\""];
   string support_url = 15 [(gogoproto.moretags) = "validate:\"omitempty,url,startsnotwith=http:,max=256\""];
   string product_url = 16 [(gogoproto.moretags) = "validate:\"omitempty,url,startsnotwith=http:,max=256\""];
+  string lsf_url = 17 [(gogoproto.moretags) = "validate:\"required_unless=LsfRevision 0,omitempty,url,startsnotwith=http:,max=256\""];
+  int32 lsf_revision = 18 [(gogoproto.moretags) = "validate:\"required_with=LsfUrl,gte=0,lte=65535\""];
```
ashcherbakov, line 42: As stated in https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/service_device_management/DistributedComplianceLedger.adoc#1512-lsfrevision

>  When the first version of the Localized String File is published, the value of this field SHOULD be 1.

So, it looks like no need to have `lsf_revision` field in `MsgCreateModel` as it must be set to `1` by DCL, doesn't it?

ankur325: OK.. then how about we remove lsf_revision field for Addition

proto/model/tx.proto
ankur325: Should we allow user to set it, or just make it auto increase whenever there is a LsfURL change.

ashcherbakov: As I understand, `lsf_revision` can be changed for the same `lsf_url`, right? This means that we should allow setting a new `lsf_revision` w/o setting/changing  `lsf_url`. DCL must check that new_lsf_revision = old_lsf_revision + 1.

In general, it looks like we have the following cases for editing LDF fields:
1) `lsf_url` is not changed ("" passed), and `lsf_revision` is changed. DCL must check that 
   - lsf_url already set for that model (non-empty strong)
   - new_lsf_revision = old_lsf_revision + 1.
2) `lsf_url` is changed (non-empty string which is different from the previous value is passed). `lsf_revision` is kept the same (0 passed). If there were no lsf for that model before (so that model's lsf_url is "" and lsf_revision is 0), then `lsf_revision` must be set to 1 by DCL.
3) Both `lsf_url` and `lsf_revision` are changed. DCL must check that new_lsf_revision = old_lsf_revision + 1.

ankur325: OK.. will update with this logic.

ashcherbakov (approved):

PR #210, opened 2022-02-02
Implemented upgrade feature

spivachuk:


x/dclupgrade/client/cli/tx_approve_upgrade.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/spf13/cobra"
+	"github.com/zigbee-alliance/distributed-compliance-ledger/utils/cli"
+	"github.com/zigbee-alliance/distributed-compliance-ledger/x/dclupgrade/types"
+)
+
+func CmdApproveUpgrade() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "approve-upgrade [name]",
+		Short: "Approve proposed upgrade with given name",
+		Args:  cobra.ExactArgs(1),
```
ashcherbakov, line 16: We usually use named flags in other commands, not arguments. Probably let's make it as in other modules (--name flag)?

x/dclupgrade/keeper/msg_server_propose_upgrade.go
```diff
@@ -0,0 +1,57 @@
...
+	// check if message creator has enough rights to propose upgrade
+	if !k.dclauthKeeper.HasRole(ctx, creatorAddr, types.UpgradeApprovalRole) {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrUnauthorized,
+			"MsgProposeUpgrade transaction should be signed by an account with the %s role",
+			types.UpgradeApprovalRole,
+		)
+	}
+
+	// check if proposed upgrade exists
+	_, isFound := k.GetProposedUpgrade(ctx, msg.Plan.Name)
```
ashcherbakov, line 28: What if we had already approved and scheduled upgrade, and then an UpgradeProposal is sent for the same name?

x/dclupgrade/handler_test.go
```diff
@@ -0,0 +1,432 @@
...
+		Handler:       NewHandler(*keeper),
+	}
+
+	return setup
+}
+
+func TestHandler_ProposedUpgradeExists(t *testing.T) {
+	setup := Setup(t)
+
+	trusteeAccAddress1 := testdata.GenerateAccAddress()
```
ashcherbakov, line 78: Why do we need this setup code in every test (lines 78-85)? Can we move it to `Setup`?

x/dclupgrade/handler_test.go
```diff
@@ -0,0 +1,432 @@
...
+	// check proposed upgrade for being created
+	proposedUpgrade, isFound := setup.Keeper.GetProposedUpgrade(setup.Ctx, msgProposeUpgrade.Plan.Name)
+	require.True(t, isFound)
+	require.Equal(t, proposedUpgrade.Plan, msgProposeUpgrade.Plan)
+	require.Equal(t, proposedUpgrade.Creator, msgProposeUpgrade.Creator)
+	require.Equal(t, proposedUpgrade.Approvals[0], msgProposeUpgrade.Creator)
+}
+
+// TODO Implement this test
+// func TestHandler_ProposeUpgradeWhenOneVoteNeeded(t *testing.T) {
```
ashcherbakov, line 178: Why can't we just `AssertCalled` with `isContextWithNonCachedMultiStore`?

x/dclupgrade/handler_test.go
```diff
@@ -0,0 +1,432 @@
...
+	setup.AddAccount(trusteeAccAddress3, []dclauthtypes.AccountRole{dclauthtypes.Trustee})
+	setup.DclauthKeeper.On("CountAccountsWithRole", mock.Anything, dclauthtypes.Trustee).Return(3)
+
+	// propose new upgrade with plan height < block height
+	msgProposeUpgrade := NewMsgProposeUpgrade(trusteeAccAddress1)
+	msgProposeUpgrade.Plan.Height = 1
+	setup.Ctx = setup.Ctx.WithBlockHeight(100)
+
+	// error returned because height in plan is less than block height on the propose stage
+	setup.UpgradeKeeper.On("ScheduleUpgrade", mock.Anything, msgProposeUpgrade.Plan).Return(sdkerrors.ErrInvalidRequest)
```
ashcherbakov, line 375: Do we have an error here just because we mocked it (`Return(sdkerrors.ErrInvalidRequest)`) or because we set `Height` less than the current block height?

x/dclupgrade/types/message_propose_upgrade_test.go
```diff
@@ -0,0 +1,154 @@
...
+		{
+			name: "invalid address",
+			msg: MsgProposeUpgrade{
+				Creator: "invalid_address",
+				Plan:    testconstants.Plan,
+			},
+			err: sdkerrors.ErrInvalidAddress,
+		},
+		{
+			name: "omitted address",
```
ashcherbakov, line 72: Should we have a test with omitted plan field?

proto/dclupgrade/tx.proto
```diff
@@ -0,0 +1,34 @@
...
+// Msg defines the Msg service.
+service Msg {
+      rpc ProposeUpgrade(MsgProposeUpgrade) returns (MsgProposeUpgradeResponse);
+  rpc ApproveUpgrade(MsgApproveUpgrade) returns (MsgApproveUpgradeResponse);
+// this line is used by starport scaffolding # proto/tx/rpc
+}
+
+message MsgProposeUpgrade {
+  string creator = 1 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.moretags) = "validate:\"required\""];
+  cosmos.upgrade.v1beta1.Plan plan = 2 [(gogoproto.nullable) = false];
```
ashcherbakov, line 20: Should we mark it as a required field?

x/dclupgrade/handler_test.go
```diff
@@ -0,0 +1,432 @@
...
+)
+
+type TestSetup struct {
+	T *testing.T
+	// Cdc         *amino.Codec
+	Ctx           sdk.Context
+	Wctx          context.Context
+	Keeper        *keeper.Keeper
+	DclauthKeeper *DclauthKeeperMock
+	UpgradeKeeper *UpgradeKeeperMock
```
ashcherbakov, line 42: It makes perfect sense to use Mocks, as we consider these tests as unit tests.
However, this module (unit) can be considered as an extension of the cosmos's upgrade module, so what if we use a real cosmos's UpgradeKeeper here? Will it simplify the tests?
For example
1) Instead of calling `AssertCalled` for `isContextWithCachedMultiStore`, it will be sufficient just to have a test where invalid Plan is proposed (if an expected error is returned, then we are sure that all needed methods have been called)
2) Instead of calling `AssertCalled` for `isContextWithNonCachedMultiStore`, we could just check that an entry in cosmos's upgrade keeper is created.

x/dclupgrade/handler_test.go
spivachuk: We define `DclauthKeeper.CountAccountsWithRole` return value right in a test. So for clarity it is better to have all the accounts being created right in a test too.

x/dclupgrade/handler_test.go
spivachuk: Because in this case `UpgradeKeeper.ScheduleUpgrade` is called with a cached context and then `writeCache` function is called.

x/dclupgrade/handler_test.go
spivachuk: `ProposeUpgrade` returns an error because `UpgradeKeeper.ScheduleUpgrade` returns it. The heights are presented here only for illustrating an example when `UpgradeKeeper.ScheduleUpgrade` returns an error.

x/dclupgrade/handler_test.go
spivachuk: As an option, tests of `x/dclupgrade` handler would be written in the described way. It might simplify the tests a bit, but would make them not pure unit tests.

docs/static/openapi.yml
```diff
@@ -3388,64 +3388,148 @@ paths:
           type: string
       tags:
         - Query
-  /dcl/model/models:
+  /zigbee-alliance/distributedcomplianceledger/dclupgrade/proposed_upgrade:
```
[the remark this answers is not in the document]
spivachuk: Fixed.

Added a corresponding information to "How To Add a new Module" section of README-DEV.

x/compliance/module.go
```diff
@@ -153,7 +153,7 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
 }
 
 // ConsensusVersion implements ConsensusVersion.
-func (AppModule) ConsensusVersion() uint64 { return 2 }
+func (AppModule) ConsensusVersion() uint64 { return 1 }
```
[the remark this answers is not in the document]
spivachuk: Added a corresponding item to "How To Add a new Module" section of README-DEV.

x/dclupgrade/client/cli/query_proposed_upgrade.go
```diff
@@ -0,0 +1,76 @@
...
+
+	flags.AddPaginationFlagsToCmd(cmd, cmd.Use)
+	flags.AddQueryFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func CmdShowProposedUpgrade() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "get-proposed-upgrade [name]",
```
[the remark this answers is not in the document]
spivachuk: Changed the CLI command to `proposed-upgrade`.

x/dclupgrade/client/cli/query_proposed_upgrade.go
```diff
@@ -0,0 +1,76 @@
...
+		Short: "Query proposed upgrade by name",
+		Args:  cobra.ExactArgs(1),
```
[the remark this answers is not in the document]
spivachuk: Done.

x/dclupgrade/client/cli/tx_approve_upgrade.go
spivachuk: Done.

x/dclupgrade/types/message_propose_upgrade_test.go
spivachuk: No, because `plan` is a struct, not a pointer to a struct.

x/dclupgrade/keeper/msg_server_propose_upgrade.go
spivachuk: We allow to create a new proposal in this case. We use the name as the unique identifier of an upgrade proposal, but as to x/upgrade module, it does not prohibit to re-schedule an upgrade with the same name multiple times.

However, in next PRs, when we introduce `ApprovedUpgrade` entity, proposing an upgrade with a previously used name should be prohibited, because we will use the name as the unique identifier of any (proposed or approved) upgrade.

ashcherbakov (approved):

PR #216, opened 2022-02-04
Enable CLI ansible role

electrocucaracha:
The command-line interface is required to operate cluster nodes. The process to install and configure it can be automated for distributed environments. This change provides ansible playbooks for performing that operation. Those playbooks are covered by Molecule unit tests which are running every time a new change is generated.


andkononykhin:
@electrocucaracha Hello. Thank you for the PR!
I need to notice though that currently we are moving towards the new non-compatible release based on the latest cosmos sdk / tendermint. It would be 0.6.0, and actually we are just cleaning up the things and will release it in a few days: code, docs, scripts are already updated. One of the changes is `dclcli` deprecation, it won't be there anymore since cosmos sdk got rid of that separation: the same binary acts as both client CLI tool and a service. The release will look like https://github.com/zigbee-alliance/distributed-compliance-ledger/releases/tag/v0.6.0-rc.2

electrocucaracha:
I wasn't aware of that depreciation, so it doesn't make sense to move forward with this change. Thanks @andkononykhin

PR #252, opened 2022-02-18
dclupgrade cli integration tests

DenisRybas:


integration_tests/cli/upgrade-demo.sh
```diff
@@ -0,0 +1,87 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+source integration_tests/cli/common.sh
+
+upgrade_height=$RANDOM
```
spivachuk, line 19: $RANDOM returns a random integer in the range from 0 to 32767. If it returns too small value then upgrade procedure can be started during a test execution. This will break the pool and thus fail the test.

So either a quite big fixed height should be used or a quite big number should be added to the random value.

integration_tests/cli/upgrade-demo.sh
```diff
@@ -0,0 +1,87 @@
...
+
+echo "propose and approve upgrade"
+random_string upgrade_name
+propose=$(dcld tx dclupgrade propose-upgrade --name=$upgrade_name --upgrade-height=$upgrade_height --upgrade-info=$upgrade_info --from alice --yes)
+approve=$(dcld tx dclupgrade approve-upgrade --name=$upgrade_name --from bob --yes)
+proposed_dclupgrade_query=$(dcld query dclupgrade proposed-upgrade --name=$upgrade_name)
+approved_dclupgrade_query=$(dcld query dclupgrade approved-upgrade --name=$upgrade_name)
+plan_query=$(dcld query upgrade plan)
+check_response "$proposed_dclupgrade_query" "Not Found" raw
+check_response "$approved_dclupgrade_query" "\"name\":\"$upgrade_name\"" raw
```
spivachuk, line 30: In positive checks in CLI integration tests we use the default output mode (which is json). So `raw` should be removed from positive checks. But a space character should be added after the colon for matching to succeed in json mode.

integration_tests/cli/upgrade-demo.sh
```diff
@@ -0,0 +1,87 @@
...
+
+set -euo pipefail
+source integration_tests/cli/common.sh
+
+upgrade_height=$RANDOM
+random_string upgrade_info
+
+echo "propose and approve upgrade"
+random_string upgrade_name
+propose=$(dcld tx dclupgrade propose-upgrade --name=$upgrade_name --upgrade-height=$upgrade_height --upgrade-info=$upgrade_info --from alice --yes)
```
spivachuk, line 24: Response code equality to 0 must be checked for success tx commands.

integration_tests/cli/upgrade-demo.sh
```diff
@@ -0,0 +1,87 @@
...
+approve=$(dcld tx dclupgrade approve-upgrade --name=$upgrade_name --from bob --yes)
+proposed_dclupgrade_query=$(dcld query dclupgrade proposed-upgrade --name=$upgrade_name)
+approved_dclupgrade_query=$(dcld query dclupgrade approved-upgrade --name=$upgrade_name)
+plan_query=$(dcld query upgrade plan)
+check_response "$proposed_dclupgrade_query" "Not Found" raw
+check_response "$approved_dclupgrade_query" "\"name\":\"$upgrade_name\"" raw
+check_response_and_report "$plan_query" "\"name\":\"$upgrade_name\"" raw
+check_response_and_report "$plan_query" "\"height\":\"$upgrade_height\"" raw
+check_response_and_report "$plan_query" "\"info\":\"$upgrade_info\"" raw
```
spivachuk, line 33: Add more output to tests (see other CLI integration tests for an example).

integration_tests/cli/upgrade-demo.sh
DenisRybas: fixed

integration_tests/cli/upgrade-demo.sh
DenisRybas: removed

integration_tests/cli/upgrade-demo.sh
DenisRybas: added checks

integration_tests/cli/upgrade-demo.sh
```diff
@@ -0,0 +1,110 @@
...
+echo "upgrade height < current height"
+random_string upgrade_name
+height=1
+propose=$(dcld tx dclupgrade propose-upgrade --name=$upgrade_name --upgrade-height=$height --upgrade-info=$upgrade_info --from alice --yes)
+echo "propose upgrade response: $propose"
+check_response_and_report "$propose" "upgrade cannot be scheduled in the past" raw
+
+test_divider
+
+echo "PASSED"
```
spivachuk, line 110: Please add empty lines between logical blocks within test cases.
It will be easier to read logical blocks when they are not glued together.

spivachuk (approved):

PR #253, opened 2022-02-18
Fix for typo in openapi

j2fong:


docs/static/openapi.yml
```diff
@@ -5629,7 +5629,7 @@ paths:
           schema:
             type: object
             properties:
-              RevokedRootCertificates:
```
ashcherbakov, line 5632: It looks like this is a typo in .proto files (https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/proto/pki/query.proto#L181), so it needs to be fixed there,  and then re-generate code and openapi (see https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/README-DEV.md#how-to-make-changes-in-data-model-for-existing-modules).
As DCL TestNet release 0.7.0 is already issued, I think we should either leave it as-is, or plan it for the next release....

ashcherbakov:
It's not sufficient just to change the .proto and openapi files... We need to re-generate the code and openapi (see https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/README-DEV.md#how-to-make-changes-in-data-model-for-existing-modules).

PR #271, opened 2022-03-03
Refactor GitHub workflows

electrocucaracha:
The current definition of the GitHub workflows causes the unnecessary execution of jobs in some cases. This PR provides changes for helping to manage the existing GitHub workflows.

- Added [Dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates) for updating package versions.
- Enabled the [Act tool](https://github.com/nektos/act) for the execution of the GitHub actions locally, e.g. `act -j check-unit-tests`
- Removed **License** and **Build** GitHub jobs. These jobs are consuming resources without providing a valid output.
- Used the [gotestfmt binary](https://debugged.it/projects/gotestfmt/) to visualize better the unit test results.
- Fixed several golang linting issues (unparam, nolintlint, thelper, nlreturn, revive, stylecheck, whitespace)


ashcherbakov:
Thank you for improving CI execution. I fully agree with the approach to not run unnecessary stuff (for example no need to run integration tests if we fixed a typo in docs).
But usually we mark some of the CI checks as required for merging a PR (it's done in the repo config, `Branch protection rule` section). Currently all checks from `verify.yml` are marked as required for `master` to not miss any checks (and it was fine as we run all the checks for all PRs).
So, now it seems we may have two problems:
- As `verify.yml` is removed in this PR, the PR can not be merged as the branch protected for `master` still expects the old checks. That's probably not a big problem, we can disable the branch protection checks for this particular PR.
- As I can see, the new approach will not run for example `Check Integration Tests` if there are no changes in the integration tests, which is fine. But how should we configure that in the `Branch protection rule` section? I don't think this config is smart enough to have such advanced policies... Probably an easy fix here would be to specify all checks (new checks) in the Branch Protection Rule (as we have it now but for new check names), so that we require all the checks to be executed, but do a change in the Job itself to not do any real work if there are no code changes according to the corresponding pattern? 
So, instead of `on/pull_request/paths`, we can consider adding `if` checks for particular jobs (https://docs.github.com/en/actions/learn-github-actions/expressions)?

electrocucaracha:
> * Branch protection rule

I see your point, maybe we can use the following two actions to have better control over the CI:

* [all-greens](https://github.com/marketplace/actions/alls-green)
* [paths-filter](https://github.com/marketplace/actions/paths-changes-filter)

I'm going to change the PR to create proposal for that idea.

.github/workflows/verify.yml
```diff
@@ -13,65 +13,75 @@
...
-  integration:
-    name: Integration tests  # remove (requires repo settings update)
+      - uses: haveyoudebuggedit/gotestfmt-action@v2
+      - name: Run tests
+        run: |
+          set -euo pipefail
+          # shellcheck disable=SC2046
+          go test -json -v $(go list ./... | grep -v '/integration_tests') 2>&1 | tee /tmp/gotest.log | gotestfmt
+  integration-tests:
+    if: needs.changes.outputs.src == 'true'
```
ashcherbakov, line 75: I believe we need to run integration tests on any changes in .go files, as well as on any changes in the integration test source code (`integration_tests` folder).

.github/workflows/verify.yml
```diff
@@ -80,42 +90,30 @@ jobs:
...
         with:
-          version: v1.43
           args: --timeout 5m0s
-
-  license:
-    name: License  # TODO remove (requires repo settings update)
+  check:
+    if: always()
+    needs:
+      - unit-tests
```
ashcherbakov, line 111: I think we need to apply the same approach then for other check: lint.yml, spell.yml, on_demand_molecule.yml, etc.
BTW superlint checks appeared to be running surprisingly long (~45 minutes). We had to disable them now, since needed to issue the release. I think we will need to optimise the super lint checks to make it much faster.

.github/workflows/verify.yml
electrocucaracha: Technically, the regex in `src` covers any change to code (go, typescript, bash, etc.) except in documentation or GitHub actions

.github/workflows/verify.yml
electrocucaracha: > I think we need to apply the same approach then for other check: lint.yml, spell.yml, on_demand_molecule.yml, etc. 

Do you mean, to put them into a single workflow file? It could be possible, but the disadvantage will be to re-run a single failure job or disable a specific workflow.

> BTW superlint checks appeared to be running surprisingly long (~45 minutes). We had to disable them now, since needed to issue the release. I think we will need to optimise the super lint checks to make it much faster.

Well, the super-linter contains a collection of linters, which combine with the size of the project as well as the variety of languages used results in longer execution. Another alternative could be to change it to run it once a week at least, in that way we can monitor any issue.

ashcherbakov: > Do you mean, to put them into a single workflow file?

I'm thinking here from the branch protection rules point of view again. It would be great to require running for example spell.yml if there are any changes in doc files, so that branch protection rules will not allow to merge a PR with failing spell checks. 
As for how to do this, put everything into a single workflow file and use all-greens action is one of the options for sure. Another option is to enable spell checks in the protection rules, but as it contains filters (paths) on the pull_request level, we should probably apply the same refactoring there as in this PR.


> Another alternative could be to change it to run it once a week at least, in that way we can monitor any issue.

It can be a reasonable option.

ashcherbakov (approved):

.github/workflows/verify.yml
```diff
@@ -13,65 +13,75 @@
...
-      - name: Set up Go 1.x
-        uses: actions/setup-go@v2
+      - uses: actions/setup-go@v2
         with:
-          go-version: ^1.16
-
-      - name: Check out code into the Go module directory
-        uses: actions/checkout@v2
-
+          go-version: ^1.13
```
ashcherbakov, line 83: Any reasons to downgrade Go version here from 1.16 to 1.13?

.github/workflows/verify.yml
ashcherbakov: BTW the `src` filter rule didn't work correctly, as OR is used in case of multiple regexps. So, it was true in most of the cases (when for example we had changes in .md files only).
It's fixed in https://github.com/zigbee-alliance/distributed-compliance-ledger/pull/300.

.github/workflows/verify.yml
electrocucaracha: I missed [this](https://github.com/zigbee-alliance/distributed-compliance-ledger/commit/8dbd9f9726e888bc986a21024f8070a29a187028) upgrade

electrocucaracha:
@ashcherbakov I submitted a fix in #310 given the issues in the last executions

PR #274, opened 2022-03-04
Abdulla ashurov/issue229 - Support read requests in load tests

uabjabborov:


bench/README.md
```diff
@@ -42,14 +41,15 @@ By that reason load test uses prepared load data which can be generated as follo
...
     python bench/generate.py bench/test.spec.yaml bench/txns
     ```
 
 Here the following (**optional**) inputs are considered:
 
 *   `NUM-USERS`: number of client accounts with write access (created as Vendors). Default: 10
 *   `NUM-REQ-PER-USER`: number of write txns to perform per a user. Default: 1000
+*   `NUM-NEW-MODELS`: models are added to the ledger for reading afterwards in load tests. Default: 5
 
 ## Run
```
ashcherbakov, line 55: It would be great to provide explicit documentation how to:
1) Run read-only requests
2) Run write-only requests
3) Run mixed mode (write and read)

ashcherbakov (approved):

PR #277, opened 2022-03-07
Add gRPC BaseQuery for /cosmos.auth.v1beta1.Query/Account

ankur325:
I was creating the views from the generated Vuex Code of Starport that uses the cosmjs library. This library for submitting a transaction does a account query that uses the base grpc query. 

https://github.com/cosmos/cosmjs/issues/1073#event-6190566653

I have added that in this PR so that the vuex actions can be dispatched.


ashcherbakov:
Will `/cosmos.auth.v1beta1.Query/Account` return our DCL Account, or Cosmos default Account?

ashcherbakov (approved):

ankur325:
> Will `/cosmos.auth.v1beta1.Query/Account` return our DCL Account, or Cosmos default Account?

It will return the base account itself, there were issues if cosmjs if I return the DCL Account.

PR #291, opened 2022-03-17
Added some clarifications to documentation

spivachuk:
- Added clarification about using dcld as a client or for initialization steps.
- Add note on adding current/bin directory to $PATH.


docs/running-node.md
```diff
@@ -107,6 +107,17 @@ curl -L -O https://raw.githubusercontent.com/zigbee-alliance/distributed-complia
 </details>
 <!-- markdownlint-enable MD033 -->
 
+> Note:
```
ashcherbakov, line 110: I think we need a similar Note in the switch_to_cosmovisor how-to docs.

spivachuk: Added

ashcherbakov (approved):

PR #302, opened 2022-03-24
Fix missing copy of cosmovisor in run_dcl_node

j2fong:


uabjabborov:
@j2fong. Can we close this pull request, since it is no longer relevant?

PR #309, opened 2022-03-28
Switch Lint workflow to schedule execution

electrocucaracha:
The execution of the Lint workflow takes ~45 mins, which causes a slow-down during the development process. Ideally, lint issues need to be addressed locally before submitting a change, but given the amount of code to validate, any small change can result in a frustrating developer experience and reduce the amount of contributions.

This change proposes to change the Lint workflow execution to weekly basis.


ashcherbakov (approved):

uabjabborov (approved):

electrocucaracha:
@ashcherbakov @uabjabborov after this change, the Lint workflow needs to be enabled.

ashcherbakov:
@electrocucaracha It's enabled now, thanks.

PR #327, opened 2022-04-11
Separate functions from the Bootstrap Ansible Role

electrocucaracha:
The `bootstrap` Ansible role performs two main functions, pulls and configures binaries and creates DCL accounts. In some cases, the accounts' creation can be performed later or skipped, so this change offers the possibility to manage it separately.


deployment/ansible/roles/configure/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: initialize the layout
+  changed_when: false
+  command: "{{ dcld.path }} init '{{ moniker | default(ansible_hostname) }}' --chain-id '{{ chain_id }}' --home {{ dcl_home }}"
+  when: not genesis_file.stat.exists
+
+- name: setup TOMLs files
+  copy:
+    dest: "{{ dcl_home }}/config/{{ item.key }}.toml"
+    content: "{{ item.value | sivel.toiletwater.to_toml  }}"
```
uabjabborov, line 29: It seems like you are replacing the whole content of a config file here. But we should keep untouched the remaining ones.

uabjabborov (approved):

PR #346, opened 2022-04-29
Abdulla ashurov/issue344

abdulla-ashurov:
Be able to re-propose Upgrade with the same name if we don't get a quorum of Trustee approvals by the specified block height (time). Now we can re-propose update


uabjabborov:
Put on hold for now

uabjabborov:
Closing for the favor of #365

PR #348, opened 2022-05-03
DCL binary install script

electrocucaracha:
Having a script that automates the DCL binary installation process can reduce time during the upgrade process and minimize the deployment errors. This PR offers an installation script to perform the installation and verification of DCL binary.

### DIFF
--- a/deployment/scripts/install.sh
+++ b/deployment/scripts/install.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+# Copyright 2022 Samsung Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


uabjabborov (approved):

deployment/scripts/install.sh
```diff
@@ -0,0 +1,91 @@
...
+        info "Downloading the DCL binary for $distro..."
+        mkdir -p "$dest"
+        url="https://github.com/zigbee-alliance/distributed-compliance-ledger/releases/download/v$version/dcld.$distro.tar.gz"
+        if command -v wget > /dev/null; then
+            wget -q -c "$url" -O - | sudo tar -xz -C "$dest"
+        elif command -v curl > /dev/null; then
+            curl -s "$url" | sudo tar -xz -C "$dest"
+        fi
+
+        sudo chown -R "$(whoami)" "$dcl_home"
```
uabjabborov, line 79: How do we ensure that `$(whoami)` is the same as the owner of `cosmovisor`? Because it is mentioned in this [doc](https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/docs/pool-upgrade-how-to.md)

electrocucaracha: Fixed on c8105b347ba421ab11cb0ed1828cc5afabf4feed

Issue #347, opened 2022-04-29
Change Vendor Account Apporval to >=1/3 Trustee approval


hawk248:
To streamline vendor account creation and approval, current 2/3+ Trustee approval is slowing vendor's ability to use testnet and will create bottleneck for production. The DCL TT decided to change it to >= 1/3 trustee approval.

No CHANGE on any other approval or revocation requirement.

abdulla-ashurov:
Plan of attack:
- Change Vendor Account Approval to >=1/3 Trustee approval.
- Add unit and integration tests (CLI, gRPC, REST API).
- Update documentation in section `Auth` for role `Vendor`.

PR:
- #351

PR #351, opened 2022-05-05
Abdulla ashurov/issue347

abdulla-ashurov:
Change Vendor Account Approval to >=1/3 Trustee approval #347

### DIFF
--- a/integration_tests/cli/auth-demo-hex.sh
+++ b/integration_tests/cli/auth-demo-hex.sh
@@ -42,51 +42,35 @@ pid_in_hex_format=0xA11
 vid=2579
 pid=2577
 
-
 echo "Jack proposes account for $user"
 result=$(echo $passphrase | dcld tx auth propose-add-account --info="Jack is proposing this account" --address="$user_address" --pubkey="$user_pubkey" --roles="Vendor" --vid="$vid_in_hex_format" --from jack --yes)
 check_response "$result" "\"code\": 0"
 
 test_divider
 
-echo "Get an account for $user is not found"
-result=$(dcld query auth account --address=$user_address)
-check_response "$result" "Not Found"
+echo "Get all active accounts. $user account in the list because has enough approvals"
+result=$(dcld query auth all-accounts)
+check_response "$result" "\"address\": \"$user_address\""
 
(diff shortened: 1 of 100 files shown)


integration_tests/cli/auth-demo-hex.sh
```diff
@@ -42,51 +42,35 @@ pid_in_hex_format=0xA11
...
+check_response "$result" "Not Found"
 
 test_divider
 
-echo "Get a proposed account for $user is not found"
-result=$(dcld query auth proposed-account --address=$user_address)
-check_response "$result" "Not Found"
+echo "Get all proposed accounts. $user account is not in the list"
+result=$(dcld query auth all-proposed-accounts)
+check_response "$result" "\[\]"
```
uabjabborov, line 73: I believe we should keep Vendor account here, because it these tests specifically check vid/pid with hex format.

integration_tests/cli/auth-demo.sh
```diff
@@ -682,4 +650,324 @@ echo "$user adds Model with VID: $vid PID: $pid"
 result=$(echo "test1234" | dcld tx model add-model --vid=$vid --pid=$pid --productName="$productName" --productLabel="Device Description" --commissioningCustomFlow=0 --deviceTypeID=12 --partNumber=12 --from=$user_address --yes 2>&1) || true
 check_response_and_report "$result" "key not found" raw
```
uabjabborov, line 652: It would be good to explain here the purpose of the new tests. i.e mentioning that vendor accounts require less quorum of trustees.

integration_tests/cli/auth-demo-hex.sh
abdulla-ashurov: Done!

abdulla-ashurov: We there keep Vendor account, only I removed tx approve-add-account, because we already have 1/3 approvals, when we propose-add-account

integration_tests/cli/auth-demo.sh
abdulla-ashurov: Done! ✅

PR #365, opened 2022-05-23
Issue 344 - Re-propose upgrade with same name

abdulla-ashurov:
- Added implementation, when proposing a new upgrade, if there is an expired upgrade proposal with the same name, we remove the proposal and add a new one (with the same name and all votes cleared) instead of failing with "Proposed upgrade with name=%v already exists on the ledger".


x/dclupgrade/keeper/msg_server_propose_upgrade.go
```diff
@@ -45,6 +33,20 @@ func (k msgServer) ProposeUpgrade(goCtx context.Context, msg *types.MsgProposeUp
 		return nil, err
 	}
 
+	// check if proposed upgrade with the same name already exists and propose plan height
+	existProposedUpgrade, isFound := k.GetProposedUpgrade(ctx, msg.Plan.Name)
```
uabjabborov, line 37: Better name would be `existingProposedUpgrade`

x/dclupgrade/keeper/msg_server_approve_upgrade.go
```diff
@@ -69,6 +69,11 @@ func (k msgServer) ApproveUpgrade(goCtx context.Context, msg *types.MsgApproveUp
 		approvedUpgrage := types.ApprovedUpgrade(proposedUpgrade)
 		k.SetApprovedUpgrade(ctx, approvedUpgrage)
 	} else {
+		if ctx.BlockHeight() > proposedUpgrade.Plan.Height {
+			return nil, sdkerrors.Wrapf(sdkerrors.ErrUnauthorized,
+				"upgrade cannot be scheduled in the past",
```
ashcherbakov, line 74:
```suggestion
				"upgrade cannot be scheduled in the past. Consider re-proposing upgrade with the same name and a new block height from the future.",
```

x/dclupgrade/keeper/msg_server_propose_upgrade.go
abdulla-ashurov: Done! ✅

x/dclupgrade/keeper/msg_server_propose_upgrade.go
```diff
@@ -45,6 +33,20 @@ func (k msgServer) ProposeUpgrade(goCtx context.Context, msg *types.MsgProposeUp
...
+	if isFound && ctx.BlockHeight() > existProposedUpgrade.Plan.Height && ctx.BlockHeight() < msg.Plan.Height {
```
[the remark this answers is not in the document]
abdulla-ashurov: Done! ✅

x/dclupgrade/keeper/msg_server_approve_upgrade.go
```diff
@@ -69,6 +69,11 @@ func (k msgServer) ApproveUpgrade(goCtx context.Context, msg *types.MsgApproveUp
...
+			)
+		}
```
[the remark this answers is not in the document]
abdulla-ashurov: Done! ✅

x/dclupgrade/keeper/msg_server_approve_upgrade.go
abdulla-ashurov: Done!

ashcherbakov (approved):

Issue #221, opened 2022-02-09
Deployment: Health&Monitoring 


ashcherbakov:
**Acceptance Criteria**
- Define ways for health and monitoring based on #100 and https://github.com/zigbee-alliance/distributed-compliance-ledger/wiki/DCL-MainNet-Deployment
- Define how to analyze node logs
  -  ELK stack?
- Define how to monitor metrics
  - Prometheus + Graphana?
- Define the metrics to be monitored
- Define metric-based alerts
- Implement scripts and/or IaC for logs
- Implement scripts and/or IaC for metrics
- Implement scripts and/or IaC for alerts

uabjabborov:
**Define how to monitor metrics:**
1. Enable Prometheus metrics in the Tendermint config 
https://docs.tendermint.com/master/nodes/metrics.html
2. For AWS use AMP (Amazon Managed Service for Prometheus) to collect Prometheus metrics and visualize using Grafana
https://aws.amazon.com/blogs/opensource/using-amazon-managed-service-for-prometheus-to-monitor-ec2-environments/

uabjabborov:
**Define how to analyze node logs:**

We can use AWS cloudwatch agent to collect `cosmovisor` service logs and it to AWS CloudWatch
https://devopscube.com/how-to-setup-and-push-serverapplication-logs-to-aws-cloudwatch/

uabjabborov:
PR #366

PR #366, opened 2022-05-23
 Deployment: Health&Monitoring #221

uabjabborov:
This PR adds IaC implementations for:
1. Cloudwatch logs for cosmovisor service
2. Prometheus metrics using AWS amp


deployment/ansible/roles/configure/vars/validator.yml
```diff
@@ -29,6 +29,8 @@ config:
     skip_timeout_commit: true
```
ashcherbakov, line 29: Should we have `skip_timeout_commit: false` (default value)?

ashcherbakov (approved):

deployment/ansible/roles/configure/vars/validator.yml
ashcherbakov: Also please make sure that https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/docs/deployment-design-aws.md contains up-to-date configs

Issue #356, opened 2022-05-11
Add "CDCertificateID" field to `DeviceSoftwareCompliance` schema, and associated query


tcarmelveilleux:
When https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/5199 lands in the spec, we would have the following changes required:

- Add mandatory `CDCertificateID` (19 character max string) to `DeviceSoftwareCompliance`
- Add a new query that can look-up all DeviceSoftwareCompliance records matching a given `CDCertificateID`

ashcherbakov:
This will be a breaking change, so better to do it before 1.0 launch.

abdulla-ashurov:
@tcarmelveilleux 
CDCertificateID is a unique key?

abdulla-ashurov:
### Plan of attack:
1. Add a new entity `DeviceSoftwareCompliance` with all fields from the entity `ComplianceInfo`.
2. Add associate query with CDCertificationID for the new entity `DeviceSoftwareCompliance`.
3. Add unit and integration tests.
4. Update documentation.

PR #377, opened 2022-06-20
Issue 356

abdulla-ashurov:
Add "CDCertificateID" field to DeviceSoftwareCompliance schema, and associated query #356


x/compliance/keeper/msg_server_provision_model.go
```diff
@@ -69,6 +69,13 @@ func (k msgServer) ProvisionModel(goCtx context.Context, msg *types.MsgProvision
 	// store compliance info
 	k.SetComplianceInfo(ctx, complianceInfo)
 
+	deviceSoftwareCompliance, _ := k.GetDeviceSoftwareCompliance(ctx, msg.CDCertificationId)
```
ashcherbakov, line 72: I think we should not update deviceSoftwareCompliance list for the provisioned state. Only for the certified state. 
So, this code can be removed.

x/compliance/keeper/msg_server_certify_model.go
```diff
@@ -97,6 +97,13 @@ func (k msgServer) CertifyModel(goCtx context.Context, msg *types.MsgCertifyMode
 	// store compliance info
 	k.SetComplianceInfo(ctx, complianceInfo)
 
+	deviceSoftwareCompliance, _ := k.GetDeviceSoftwareCompliance(ctx, msg.CDCertificationId)
```
ashcherbakov, line 100: We need to remove the compliance info from the list in case of revocation.

x/compliance/keeper/msg_server_provision_model.go
abdulla-ashurov: Done! ✅

x/compliance/keeper/msg_server_certify_model.go
abdulla-ashurov: Done! ✅

x/compliance/types/device_software_compliance.pb.go
```diff
@@ -0,0 +1,400 @@
...
+			return 0, ErrInvalidLengthDeviceSoftwareCompliance
+		}
+		if depth == 0 {
+			return iNdEx, nil
+		}
+	}
+	return 0, io.ErrUnexpectedEOF
+}
+
+func (m *DeviceSoftwareCompliance) IsComplianceInfoExist(
```
ashcherbakov, line 380: Never put any custom logic to .pb.go files as they can be re-generated! Please move  `IsComplianceInfoExist` and `RemoveComplianceInfo` to `keeper/device_software_compliance.go` file.

abdulla-ashurov: Done! ✅

x/compliance/keeper/device_software_compliance.go
```diff
@@ -60,3 +60,19 @@ func (k Keeper) GetAllDeviceSoftwareCompliance(ctx sdk.Context) (list []types.De
 
 	return
 }
+
+func IsComplianceInfoExist(
```
ashcherbakov, line 64: If can make it a method of `DeviceSoftwareCompliance`. Just create a file `x/compliance/types/device_software_compliance.go` in addition to `x/compliance/types/device_software_compliance.pb.go` . See for example `x/compliance/types/compliance_info.go`.

ashcherbakov (approved):

PR #433, opened 2023-02-20
Issue 429 optional fields

lazizcodesatdsr:
Made `partNumber` and `productLabel` optional fields.


ashcherbakov (approved):
LGTM. Just make sure that all checks in the PR are green.

ashcherbakov (approved):
LGTM.
We may rethink which linters should be enabled and which disabled. Let's do it as a separate PR/task.

Issue #421, opened 2022-11-11
Deletion of Model Versions


ashcherbakov:
There is a command to delete a model, but there is no command to delete a model version.

1.  Model version can be deleted by the owner Vendor (account with the VID created the model) only before it is certified and certification is issued by CSA.
2. When a Model is deleted, the corresponding model versions must be deleted as well
3. Model can be deleted by the owner Vendor (account with the VID created the model) only before at least one of its Model Versions is certified and certification is issued by CSA.
   - Example: (A)/(B)/1.0, (A)/(B)/2.0, (A)/(B)/3.0 (Certified). In this scenario, ModelVersion (A)/(B)/(1.0 - 2.0) can be deleted. However, model (A)/(B) can not be deleted because of the certification record on (A)/(B)/3.0

  
Notes:
- The deletion must be a ~~soft-delete~~ final delete (completely remove from the State the same way as a Model)
- ~~GET_DELETED_MODEL_VERSION(s) needs to be implemented~~
- Separate PRs cane be sent for each item (1, 2, 3)
- The model/modelversion can not be deleted, if its certification status is revoked or provisioned. In other words, deletion can not happen if there is any certification status.

lazizcodesatdsr:
FYI, @DenisRybas has initialized the deletion (https://github.com/zigbee-alliance/distributed-compliance-ledger/tree/delete-model-version)
While he is busy with another task, I will take the implementation for the point 1.

PR #454, opened 2023-03-22
Delete model with associated model versions

lazizcodesatdsr:
Point 2 at https://github.com/zigbee-alliance/distributed-compliance-ledger/issues/421:
> When a Model is deleted, the corresponding model versions must be deleted as well


x/model/handler_test.go
```diff
@@ -427,6 +427,53 @@ func TestHandler_DeleteModel(t *testing.T) {
 	require.Equal(t, codes.NotFound, status.Code(err))
 }
 
+func TestHandler_DeleteModelWithAssociatedModelVersions(t *testing.T) {
```
DenisRybas, line 430: Please also add integration tests (GRPC, REST, CLI) for it

x/model/keeper/msg_server_model.go
```diff
@@ -195,5 +200,31 @@ func (k msgServer) DeleteModel(goCtx context.Context, msg *types.MsgDeleteModel)
 	// remove product from VendorProducts
 	k.RemoveVendorProduct(ctx, msg.Vid, msg.Pid)
 
+	modelVersions, err := k.ModelVersions(goCtx, &types.QueryGetModelVersionsRequest{
```
DenisRybas, line 203: I think it will be cleaner to use `k.GetModelVersions()`, because this code fragment will not differ from others in approach

x/model/keeper/msg_server_model.go
```diff
@@ -195,5 +200,31 @@ func (k msgServer) DeleteModel(goCtx context.Context, msg *types.MsgDeleteModel)
...
+		Vid: msg.Vid, 
+		Pid: msg.Pid,
+	})
+
+	if err != nil && status.Code(err) == codes.InvalidArgument {
```
DenisRybas, line 208: if the approach mentioned above is used, the return value `found` can be used here

x/model/keeper/msg_server_model.go
```diff
@@ -195,5 +200,31 @@ func (k msgServer) DeleteModel(goCtx context.Context, msg *types.MsgDeleteModel)
...
+		return nil, err
+	} 
+
+	if modelVersions != nil {
```
DenisRybas, line 212: if the approach mentioned above is used, this `if` may be removed

DenisRybas (changes requested):
Thanks for your work! I left a couple of minor comments

x/model/keeper/msg_server_model.go
lazizcodesatdsr: @DenisRybas ah, I missed this function, thanks!

DenisRybas (approved):
LGTM. Thank you!

Issue #458, opened 2023-03-23
DA PKI Revocation: Design


ashcherbakov:
**User stories:**
- Revocation of PAI (by proof of key possession): Member who can prove ownership of PAA private key can submit a transaction to publish the location of the CRL distribution point for that PAA. This information can be updated by the member using the same proof.
 - Revocation of a DAC (by proof of key possession): Member who can prove ownership of PAI private key can submit a transaction to publish the location of the CRL distribution point for that PAI. This information can be updated by the member using the same proof.
 -  Revocation of PAI  (by Trustees): A Trustee can propose a transaction to publish the location of the CRL distribution point for a PAA. This SHALL require approval of two thirds of the trustees vote. 
 - Revocation of DAC (by Trustees): A Trustee can propose a transaction to publish the location of the CRL distribution point for a PAI. This SHALL require approval of two thirds of the trustees vote. 
 
**Notes**:
- PAA = Root certificate
- PAI = Intermediate certificate
- DAC = Leaf certificate
- CRL = Certificate Revocation List

**Acceptance criteria**:
Create a design doc that addresses the following:
  - Design the API, model and state (key-value)
  - Design Auth rules
  - Consider a model from https://csamembers.slack.com/archives/C025PQ6RMRA/p1678980910132059
  - Think if `subject` is needed (we use `subject` for all certificate models, but the proposal above has only `subjectKeyID`)  
  - Think if we need to validate `vid` field somehow 
  - Think what GET (Read) API we need. Think if we need some additional indices in the state for this.
  - Think if existing models/states need to be adjusted. Think if we need to update existing entires in the Main Net (and hence migration of data is needed).
  - If migration of data is needed, think how it should be done via standard Cosmos/cosmovisor/upgrade module means.
     - See https://docs.cosmos.network/main/core/upgrade  
  - How this new functionality should affect existing certificate "revocation" logic of DCL. For example, if a certificate is revoked (by the current DCL means), should the corresponding CRL endpoints be removed?
  - Think how to properly do "proof of possession" stories. Usually some "nonce" needs to be used to avoid replay attacks, so the DCL API needs to designed correspondingly (DCL: generate nonce -> User: sign nonce by the key -> DCL: verify signature)
  - Consider reading the RFC 5280 CRL spec: https://www.ietf.org/rfc/rfc5280.txt

PR #472, opened 2023-04-28
DA PKI Revocation design

ashcherbakov:
DA PKI Revocation Design based on https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/6360 and discussions on DCL TT group.
Addresses #458.

### DIFF
--- a/docs/design/da-pki-revocation-design.md
+++ b/docs/design/da-pki-revocation-design.md
@@ -1,0 +1,213 @@
+# DA PKI Revocation
+
+## User Stories
+
+1. Revocation of PAI
+
+   A Vendor with DCL write privilege can submit a transaction to publish the location of the CRL distribution point for a PAA associated with the Vendor ID. This information can be updated by the Vendor.
+
+2. Revocation of a DAC 
+
+   A Vendor with DCL write privilege can submit a transaction to publish the location of the CRL distribution point for a PAI associated with the Vendor ID. This information can be updated by the Vendor.
+
+
+## Revocation Distribution Point Schema
+The schema is the same as in [PR360](https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/6360).
+However, the validation logic is proposed to be a bit different (see [Validation Logic](#validation-logic) and [Divergence](#divergence-from-pr360httpsgithubcomchip-specificationsconnectedhomeip-specpull6360)). 
+
+
+| Name                 | Type    | Constraint          | Mutable | Conformance |
+|:---------------------|:--------|:--------------------|:--------|-------------|
+| VendorID             | uint16  | all                 | No      | M           |
+| ProductID            | uint16* | all                 | No      | O (desc)    |
+| IsPAA                | bool    | all                 | No      | M           |
+| Label                | string  | max 64              | No      | M           |
+| CRLSignerCertificate | string  | max 2048            | No      | M           |
+| IssuerSubjectKeyID   | string  | max 64              | No      | M           |
+| DataUrl              | string  | max 256             | Yes     | M           |
+| DataFileSize         | uint64  | desc                | Yes     | O           |
+| DataDigest           | string  | max 128             | Yes     | O (desc)    |
+| DataDigestType       | uint32  | all                 | Yes     | O (desc)    |
+| RevocationType       | uint32  | all                 | No      | M           |
+
+*: [PR360](https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/6360) defines PID as `uint32` which is incorrect. PID must be `uint16` as in other places.
+
+## Divergence from [PR360](https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/6360)
+1. No validation of `DataUrl` content.
+   If the content is invalid, it can be edited/removed by either owner Vendor or Trustees via Update and Delete commands.
+
+   Such validation requires resolving an external URL, and this is a non-deterministic operation:
+   some nodes may resolve the URL successfully, but some nodes may not resolve it or be redirected to a different content.
+   Non-deterministic results of validation may cause a situation that on some nodes the transaction is considered as valid,
+   and on some nodes as invalid. That will break the consensus protocol and DCL won't be able to process write requests anymore.
+   An alternative option is to pass the whole content of `DataUrl` to the transaction, but it doesn't make a lot of sense because
+    1. the content can be quite big (1 MB)
+    2. there will be no guarantee that `DataUrl` actually points to the same content as passed
+
+2. Validation of `CRLSignerCertificate`. If `CRLSignerCertificate` is a PAA, then it must be present on DCL (there must be a PAA on DCL with the same pem value).
+   If `CRLSignerCertificate` is a PAI, then it must be chained back to a valid PAA present on the ledger.
+
+3. If `IssuerSubjectKeyID` is not equal to the `Subject Key Identifier` of `CRLSignerCertificate`, the transaction is considered invalid and rejected.
+
+4. If `RevocationType` is 1 (RFC5280 CRL), then `DataFileSize`, `DataDigest`, `DataDigestType` must be empty.
+
+   Otherwise, it may cause confusion. Certificates to be revoked are added to the content of CRL URL without a need to send a transaction to the ledger.
+   So, someone may look at the DCL entry, see that a file size is specified, then go to the URL, and see that the actual size there is different.
+   It may cause a question if the data is valid, and why a different file size is in DCL.
+
+5. `ProductID` is expected to be `unit16`, not `unit32` as in other places.
+
+
+## A need for Proof-of-possession of `CRLSignerCertificate` key
+No additional proof-of-possession is required because 
+- The transaction is already signed by the Vendor account key, so only a trusted Vendor can send it.
+- DCL checks that `CRLSignerCertificate` is a valid PAI (signed by a trusted PAA from DCL) or a PAA present on the ledger.
+- It can be complicated to create an interactive PoP of `CRLSignerCertificate` for some PAIs.
+- Indirect CRL are out of scope for now
+
+## Transactions
+
+### 1. ADD_PKI_REVOCATION_DISTRIBUTION_POINT
+Publishes a PKI Revocation distribution endpoint owned by the Vendor. 
+
+If `crlSignerCertificate` is a PAA (root certificate), then it must be present on DCL.
+
+If `crlSignerCertificate` is a PAI (intermediate certificate), then it must be chained back to a valid PAA (root certificate) present on DCL.
+In this case `crlSignerCertificate` is not required to be present on DCL, and will not be added to DCL as a result of this transaction.
+If PAI needs to be added to DCL, it should be done via `ADD_X509_CERT` transaction.
+
+Publishing the revocation distribution endpoint doesn't automatically remove PAI (Intermediate certificates)
+and DACs (leaf certificates) added to DCL if they are revoked in the CRL identified by this distribution point.
+
+
+- Who can send:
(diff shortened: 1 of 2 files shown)


DenisRybas (approved):
LGTM

docs/design/da-pki-revocation-design.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+| Name                 | Type    | Constraint          | Mutable | Conformance |
+|:---------------------|:--------|:--------------------|:--------|-------------|
+| VendorID             | uint16  | all                 | No      | M           |
+| ProductID            | uint16* | all                 | No      | O (desc)    |
+| IsPAA                | bool    | all                 | No      | M           |
+| Label                | string  | max 64              | No      | M           |
+| CRLSignerCertificate | string  | max 2048            | No      | M           |
+| IssuerSubjectKeyID   | string  | max 64              | No      | M           |
```
tcarmelveilleux, line 26: This may need to become mutable, and some checks done on transcations where an update is done. Would that work?

docs/design/da-pki-revocation-design.md
```diff
@@ -0,0 +1,213 @@
...
+   Such validation requires resolving an external URL, and this is a non-deterministic operation:
+   some nodes may resolve the URL successfully, but some nodes may not resolve it or be redirected to a different content.
+   Non-deterministic results of validation may cause a situation that on some nodes the transaction is considered as valid,
+   and on some nodes as invalid. That will break the consensus protocol and DCL won't be able to process write requests anymore.
+   An alternative option is to pass the whole content of `DataUrl` to the transaction, but it doesn't make a lot of sense because
+    1. the content can be quite big (1 MB)
+    2. there will be no guarantee that `DataUrl` actually points to the same content as passed
+
+2. Validation of `CRLSignerCertificate`. If `CRLSignerCertificate` is a PAA, then it must be present on DCL (there must be a PAA on DCL with the same pem value).
+   If `CRLSignerCertificate` is a PAI, then it must be chained back to a valid PAA present on the ledger.
```
tcarmelveilleux, line 48: - Chaining should be validated via:
  - Query for a PAA where `Subject == CRLSignerCertificate.Issuer` and `SubjectKeyID == CRLSignerCertificate.AuthorityKeyId"
  - Build certification path with both elements, verify path.

docs/design/da-pki-revocation-design.md
```diff
@@ -0,0 +1,213 @@
...
+
+3. If `IssuerSubjectKeyID` is not equal to the `Subject Key Identifier` of `CRLSignerCertificate`, the transaction is considered invalid and rejected.
+
+4. If `RevocationType` is 1 (RFC5280 CRL), then `DataFileSize`, `DataDigest`, `DataDigestType` must be empty.
+
+   Otherwise, it may cause confusion. Certificates to be revoked are added to the content of CRL URL without a need to send a transaction to the ledger.
+   So, someone may look at the DCL entry, see that a file size is specified, then go to the URL, and see that the actual size there is different.
+   It may cause a question if the data is valid, and why a different file size is in DCL.
+
+5. `ProductID` is expected to be `unit16`, not `unit32` as in other places.
```
tcarmelveilleux, line 58: Will do in a follow-up in spec.

docs/design/da-pki-revocation-design.md
```diff
@@ -0,0 +1,213 @@
...
+If PAI needs to be added to DCL, it should be done via `ADD_X509_CERT` transaction.
+
+Publishing the revocation distribution endpoint doesn't automatically remove PAI (Intermediate certificates)
+and DACs (leaf certificates) added to DCL if they are revoked in the CRL identified by this distribution point.
+
+
+- Who can send:
+    - Vendor account
+    - `vid` field in the transaction (`VendorID`) must be equal to the Vendor account's VID
+    - `vid` field in the `CRLSignerCertificate` (for vendor-scoped PAAs and PAIs) must be equal to the Vendor account's VID
```
tcarmelveilleux, line 86: Additional rules:

- PAA revocation point
  - If VID is present in the PAA certificate already in DCL, after finding the existing PAA certificate then `vid` field must match that --> Case of Vendor-Scoped PAA.
  - For non-vendor-scoped PAA, `vid` must match the account `vid`  associated with the PAA proposal for the existing matching PAA found.
    - `vid` must be present in  `CRLSignerCertificate` issuer

PAI revocation point:
  - If VID is present in the PAA certificate already in DCL, after finding the existing PAA certificate then `vid` field must match that --> Case of Vendor-Scoped PAA.
  - For non-vendor-scoped PAA, `vid` must match the account `vid`  associated with the PAA proposal for the existing matching PAA found.
- `vid` must be present in  `CRLSignerCertificate` issuer

In all VID/PID check cases, the rules for extracting VID/PID from subject/issuer fields are in `6.2.2.2. Encoding of Vendor ID and Product ID in subject and issuer fields` --> both OID and text versions supported.

docs/design/da-pki-revocation-design.md
```diff
@@ -0,0 +1,213 @@
...
+- Parameters:
+    - vid: `uint16` -  Vendor ID (positive non-zero)
+    - pid: `optional(uint16)` -  Product ID (positive non-zero)
+    - isPAA: `bool` -  True if the revocation information distribution point relates to a PAA
+    - label: `string` -  A label to disambiguate multiple revocation information partitions of a particular issuer.
+    - crlSignerCertificate: `string` -  PEM encoded certificate (string or path to file containing data)
```
tcarmelveilleux, line 92: Always the actual certificate. Not a path to file containing data.

docs/design/da-pki-revocation-design.md
```diff
@@ -0,0 +1,213 @@
...
+    - issuerSubjectKeyID: `string` -  crlSignerCertificate's `Subject Key Id` in hex string format, e.g: `5A:88:0E:6C:36:53:D0:7F:B0:89:71:A3:F4:73:79:09:30:E6:2B:DB`
```
tcarmelveilleux, line 93: Spec text is clear about explicit format:

```
For example, subject key ID A3:03:13:6D:54:A8:4B:E2:4C:48:87:B3:41:06:6D:C2:70:96:2F:99 (as it would appear in openssl x509 output, for human consumption) would be recorded as A303136D54A84BE24C4887B341066DC270962F99.
```

This does not exactly match the DCL PAA schema. Partly this is because the PAA DCL schema was "back-ported" after implementation (which made particular choices) back to spec, and the Core spec has a cross-reference to DCL implementation, which is problematic since it can move.

We have to resolve this.

The method used here (no colons) can easily renormalized for equality check with approved PAAs schema.

docs/design/da-pki-revocation-design.md
```diff
@@ -0,0 +1,213 @@
...
+    - dataUrl: `string` -  The URL where to obtain the information in the format indicated by the RevocationType field
+    - dataFileSize: `optional(uint64)` -  Total size in bytes of the file found at the DataUrl. Must be omitted if RevocationType is 1.
+    - dataDigest: `optional(string)` -  Digest of the entire contents of the associated file downloaded from the DataUrl. Must be omitted if RevocationType is 1.
+    - dataDigestType: `optional(uint32)` - The type of digest used in the DataDigest field from the list of [1, 7, 8, 10, 11, 12] (IANA Named Information Hash Algorithm Registry).
+    - revocationType: `uint32` - The type of file found at the DataUrl for this entry. Supported types: 1 - RFC5280 Certificate Revocation List (CRL).
+- In State:
+    - `pki/RevocationDistributionPoint/value/<IssuerSubjectKeyID>` -> list of Revocation Distribution Points
+    - `pki/RevocationDistributionPoint/value/<IssuerSubjectKeyID>/<vid>/<label>`-> Revocation Distribution Point
```
tcarmelveilleux, line 101: Why is the vid included in this query?

You could not have the same IssuerSubjectKeyID for a different vid.

docs/design/da-pki-revocation-design.md
ashcherbakov: What is the use case to mutate it?
I thought that we have an assumption that `IssuerSubjectKeyID` is equal to the CRLSignerCertificate's Subject Key Identifier. What are the cases when they are different?

docs/design/da-pki-revocation-design.md
ashcherbakov: Yes, this is exactly what's meant here. I will add the clarification.

docs/design/da-pki-revocation-design.md
```diff
@@ -0,0 +1,213 @@
...
+## Transactions
+
+### 1. ADD_PKI_REVOCATION_DISTRIBUTION_POINT
+Publishes a PKI Revocation distribution endpoint owned by the Vendor. 
+
+If `crlSignerCertificate` is a PAA (root certificate), then it must be present on DCL.
+
+If `crlSignerCertificate` is a PAI (intermediate certificate), then it must be chained back to a valid PAA (root certificate) present on DCL.
+In this case `crlSignerCertificate` is not required to be present on DCL, and will not be added to DCL as a result of this transaction.
+If PAI needs to be added to DCL, it should be done via `ADD_X509_CERT` transaction.
```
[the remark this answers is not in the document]
ashcherbakov: I think it's not obvious (at least from the DCL API point of view) that this PAI is not added to DCL (especially since we require that PAA must be on DCL) . I propose to keep this statement to avoid further questions and confusion.

docs/design/da-pki-revocation-design.md
ashcherbakov: Yes, these rules are already assumed and mentioned in the `Validation Logic` section.

ashcherbakov: > In all VID/PID check cases, the rules for extracting VID/PID from subject/issuer fields are in 6.2.2.2. Encoding of Vendor ID and Product ID in subject and issuer fields --> both OID and text versions supported.


Yes, this is exactly how it's assumed to be processed. I will add a clarification to make it more clear.

docs/design/da-pki-revocation-design.md
ashcherbakov: Yes, the actual transaction will always consist the actual certificate. We are going to support path for the corresponding parameter in CLI (as we do with other X509 related CLI tins). I will update it to avoid confusion.

docs/design/da-pki-revocation-design.md
ashcherbakov: The [spec](https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/service_device_management/DistributedComplianceLedger.adoc#84-label) says that `Uniqueness within the Device Attestation PKI Revocation Distribution Points schema SHALL be enforced against the tuple containing all of: (VendorID, Label, IssuerSubjectKeyID)`.

Right now we don't have any specific validation rules that that `You could not have the same IssuerSubjectKeyID for a different vid`. It seems usually it's the case because we check that PAIs are chained back to valid PAAs, so it's unlikely that there two valid PAIs with the same `IssuerSubjectKeyID` and different `vid`.
But for non-VID scoped PAA it can be the case (see https://github.com/zigbee-alliance/distributed-compliance-ledger/pull/472/files#r1186284795, that any Vendor can propose a Revocation endpoint for non-vid scoped PAAs, as was agreed on DCL TT call).

docs/design/da-pki-revocation-design.md
ashcherbakov: Yes, thanks for noticing, I will fix it.
As of now, in order to query certificates in DCL, the SKID is expected to be with colons. So, it seems the `issuerSubjectKeyID` value will need to be split by colons, and then this new value can be used for searching PAAs on DCL as part of the checks.

PR #514, opened 2023-08-21
Compliance module unit tests refactoring

DenisRybas:


x/compliance/handler_update_compliance_info_test.go
```diff
@@ -0,0 +1,205 @@
...
+	parentChild := testconstants.ParentChild1
+	isUpdatedMinimally := true
+
+	_, updateComplianceInfoErr := setup.UpdateComplianceInfoOneOptionalField(vid, pid, softwareVersion, softwareVersionString, certificationType, parentChild, setup.CertificationCenter)
+	require.NoError(t, updateComplianceInfoErr)
+
+	updatedComplianceInfo := queryExistingComplianceInfo(setup, vid, pid, softwareVersion, certificationType)
+	updatedDeviceSoftwareCompliance := queryExistingDeviceSoftwareCompliance(setup, cDCertificateID)
+
+	require.NotEqual(t, originalComplianceInfo.ParentChild, updatedComplianceInfo.ParentChild)
```
ashcherbakov, line 147: It's not clear from reading the test case why do we check for ParentChild here

x/compliance/keeper/msg_server_update_compliance_info.go
```diff
@@ -104,6 +104,10 @@ func (k msgServer) UpdateComplianceInfo(goCtx context.Context, msg *types.MsgUpd
 		complianceInfo.Transport = msg.Transport
 	}
 
+	if msg.Owner != "" {
```
ashcherbakov, line 107: It seems it was an overlook to add Owner to Update Compliance .proto....
It should not be editable. 
I propose to leave the Owner field non-editable (do not include this code), and rather remove the Owner field from the documentation.

akarabashov (approved):

Issue #523, opened 2023-11-10
PID scoped Accounts


ashcherbakov:
Provide account access keys that are scoped to a range of product ID’s (PIDs).
Currently account access keys are scoped to vendor ID’s (VID) only. See
- https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/proto/dclauth/account.proto#L22
- https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/x/model/keeper/check_model_rights.go#L9

If a PID is specified for the account, then the account can manage Model/ModelVersion with the given PID only (similar to VID scoped Accounts).

**Example:**
- Vendor Account Key 1: 0x ABC…. , VID = 0x 1111, PID [0000 - 0011] (OPTIONAL)
- Vendor Account Key 2: 0x DEF…. , VID = 0x 1111, PID [0012 - 1111] (OPTIONAL)
- Vendor Account Key 3: 0x AEF…. , VID = 0x 1111 
Once approved, Key1 will only be able to create/update records (model/modelversions) with VID = 0x 1111, PID [0000 - 0011]. Key3 can create/update all records (model/modelversions). If the PID ranges are submitted, they are enforced to be non-overlapping.

**Acceptance Criteria:**
- Extend the Account with a new optimal field(s)
  - modify .proto files (https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/proto/dclauth/account.proto#L22)
  - regenerate the code, see https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/README-DEV.md#how-to-make-changes-in-data-model-for-existing-modules
- Extend the business logic (https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/x/model/keeper/check_model_rights.go ?) to check for PIDs when creating/updating/deleting a Model or ModelVersion.
- Add Unit Tests. Consider improving/simplifying existing unit tests if it makes sense. Create tests for both positive and negative scenarios. 
  - https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/x/model/handler_test.go
  - https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/x/model/types/messages_model_test.go
  - https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/x/model/types/messages_model_version_test.go 
- Add Integration tests (CLI and REST). Positive and basic negative tests (to check that errors are returned) are enough.
[...]

Issue #519, opened 2023-09-07
[DCL REST API] Query the Matter PAA certificate with only subjectKeyId


wqx6:
Now we can query a certain Matter PAA certificate with DCL REST APIs if we have both subject and subjectKeyId or only subject of that PAA certificate.
```
curl -X 'GET' 'https://on.dcl.csa-iot.org/dcl/pki/certificates/{subject}/{subjectKeyId}'

curl -X 'GET' 'https://on.dcl.csa-iot.org/dcl/pki/certificates/{subject}'
```
Could we have a DCL REST API that can be used to query PAA certificate with only subjectKeyId? Like this:
```
curl -X 'GET' 'https://on.dcl.csa-iot.org/dcl/pki/certificates?subjectKeyId={subjectKeyId}'
```

jcps07:
DCL-TT: Recommended for 1.3

ashcherbakov:
What is important here:
We will need to write a migration script for existing PAAs (on TestNet and MainNet) which will be executed during upgrade. The migration script should construct the new index to allow querying by SKID.

akarabashov:
PR: https://github.com/zigbee-alliance/distributed-compliance-ledger/issues/519

akarabashov:
- Functionality implemented as described in the ticket
- Unit and integration tests are added
- Implemented Store Migration
- Docs are updated

wqx6:
@ashcherbakov  Hi, is the new REST API of querying PAA with SKID available now? If yes, how can I use the API. I didn't see update in https://on.dcl.csa-iot.org/

ashcherbakov:
It's implemented, but not deployed to TestNet and MainNet yet.
You can look at the API at https://zigbee-alliance.github.io/distributed-compliance-ledger/

PR #532, opened 2024-01-10
#519 Query Certificates with subjectKeyId

Abdulbois:
- Enable querying certs with subjectKeyId
- Add migration flow (**pki** module - from 2 to 3 version)
- Update OpenApi yaml file (includes missed changes from #523)


x/pki/keeper/msg_server_approve_add_x_509_root_cert.go
```diff
@@ -89,6 +89,9 @@ func (k msgServer) ApproveAddX509RootCert(goCtx context.Context, msg *types.MsgA
 
 		// add to subject -> subject key ID map
 		k.AddApprovedCertificateBySubject(ctx, rootCertificate.Subject, rootCertificate.SubjectKeyId)
+
+		// add to subject key ID -> certificates map
+		k.AddApprovedCertificateBySubjectKeyID(ctx, rootCertificate)
```
akarabashov, line 94: Please add a test to 'handler_test.go' to ensure that this method correctly modifies this state.

x/pki/keeper/msg_server_approve_revoke_x_509_root_cert.go
```diff
@@ -75,6 +75,9 @@ func (k msgServer) ApproveRevokeX509RootCert(goCtx context.Context, msg *types.M
 
 		// remove from subject -> subject key ID map
 		k.RemoveApprovedCertificateBySubject(ctx, msg.Subject, msg.SubjectKeyId)
+
+		// remove from subject key ID -> certificates map
+		k.RemoveApprovedCertificatesBySubjectKeyID(ctx, msg.SubjectKeyId)
```
akarabashov, line 80: Please add a test to 'handler_test.go' to ensure that this method correctly modifies this state

x/pki/keeper/msg_server_add_x_509_cert.go
```diff
@@ -99,5 +99,8 @@ func (k msgServer) AddX509Cert(goCtx context.Context, msg *types.MsgAddX509Cert)
 	// add to subject -> subject key ID map
 	k.AddApprovedCertificateBySubject(ctx, certificate.Subject, certificate.SubjectKeyId)
 
+	// add to subject key ID -> certificates map
+	k.AddApprovedCertificateBySubjectKeyID(ctx, certificate)
```
akarabashov, line 103: Please add a test to 'handler_test.go' to ensure that this method correctly modifies this state

x/pki/keeper/msg_server_revoke_x_509_cert.go
```diff
@@ -49,5 +49,8 @@ func (k msgServer) RevokeX509Cert(goCtx context.Context, msg *types.MsgRevokeX50
 	// remove from subject -> subject key ID map
 	k.RemoveApprovedCertificateBySubject(ctx, msg.Subject, msg.SubjectKeyId)
 
+	// remove from subject key ID -> certificates map
+	k.RemoveApprovedCertificatesBySubjectKeyID(ctx, msg.SubjectKeyId)
```
akarabashov, line 53: Please add a test to 'handler_test.go' to ensure that this method correctly modifies this state

akarabashov (commented):
Please update 'docs/transactions.md'

x/pki/keeper/approved_certificates_test.go
```diff
@@ -22,6 +22,10 @@ func createNApprovedCertificates(keeper *keeper.Keeper, ctx sdk.Context, n int)
 		items[i].SubjectKeyId = strconv.Itoa(i)
 
 		keeper.SetApprovedCertificates(ctx, items[i])
+		keeper.SetApprovedCertificatesBySubjectKeyID(ctx, types.ApprovedCertificatesBySubjectKeyId{
+			SubjectKeyId: items[i].SubjectKeyId,
+			Certs:        items[i].Certs,
+		})
```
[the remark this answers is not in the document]
Abdulbois: See x/pki/keeper/grpc_query_approved_certificates_test.go:125 test, it is needed to check retrieving certificates by `subject-id`

docs/transactions.md
```diff
@@ -1245,7 +1245,7 @@ If a Revocation Distribution Point (such as RFC5280 Certificate Revocation List)
 
 **Status: Implemented**
```
ashcherbakov, line 1247: I would add a new section GET_X509_CERTS_BY_SUBJECT_KEY_ID instead of modifying existing one to make it clear that we support querying by SUBJECT_KEY_ID.
- CLI command: `dcld query pki x509-cert --subject-key-id=<hex string>`
- REST:  `/dcl/pki/certificates?subjectKeyId={subjectKeyId}`

integration_tests/run-all.sh
```diff
@@ -131,12 +131,12 @@ if [[ $TESTS_TO_RUN =~ "all" || $TESTS_TO_RUN =~ "light" ]]; then
   CLI_SHELL_TESTS=$(find integration_tests/light_client_proxy -type f -name '*.sh' -not -name "common.sh")
```
ashcherbakov, line 132: Please add integration tests for query certs by subject-key-id (CLI, REST, gRPS)

x/pki/client/cli/query_approved_certificates.go
```diff
@@ -12,6 +12,7 @@ import (
 )
```
ashcherbakov, line 13: 1) I would keep `all-x509-certs` untouched (so that it returns all certs with pagination as other `all-xxx` commands).
I would update the `x509-cert` command to make `subject` optional there, so that if subject is specified - existing logic is called (get certificates by subject+subject-key-id), if `subject` is not specified - then all certificates by subject-key-id are returned.
2) Please make sure that get by subject-key-id works with light client approach by using `cli.QueryWithProof` instead of `types.QueryAllApprovedCertificatesRequest`.

integration_tests/upgrade/test-upgrade-1.2-to-1.3.sh
```diff
@@ -0,0 +1,1660 @@
...
+echo "Get all revoked x509 root certificates"
+result=$($DCLD_BIN_NEW query pki all-revoked-x509-root-certs)
+check_response "$result" "\"subject\": \"$root_cert_subject_new\""
+check_response "$result" "\"subjectKeyId\": \"$root_cert_subject_key_id_new\""
+
+echo "Get all proposed x509 root certificates to revoke"
+result=$($DCLD_BIN_NEW query pki all-proposed-x509-root-certs-to-revoke)
+check_response "$result" "\"subject\": \"$test_root_cert_subject_new\""
+check_response "$result" "\"subjectKeyId\": \"$test_root_cert_subject_key_id_new\""
+
```
ashcherbakov, line 1595: Please make sure that the new call (`x509-cert --subject-key-id=...`) works after upgrade

x/pki/handler_test.go
```diff
@@ -412,11 +412,20 @@ func TestHandler_ApproveAddX509RootCert_ForNotEnoughApprovals(t *testing.T) {
 
 	// query approved certificate and we should get one back
```
ashcherbakov, line 413: Do we have a unit test where there are multiple certificate with the same SubjectKeyId (but different subjects) are added? We need to check that they all are correctly returned (not overridden).
Also, we need to check that if a certificate with subject1-subject-key-id2 is revoked, subject2-subject-key-id is not removed and can still be queried by subject-key-id.

x/pki/keeper/msg_server_revoke_x_509_cert.go
ashcherbakov, line 53: What if we have multiple certificates with the same SubjectKeyId, but different subjects? It seems they all will be removed, which is not correct (here and in other similar places).

ashcherbakov (changes requested):

integration_tests/upgrade/test-upgrade-1.2-to-1.3.sh
```diff
@@ -0,0 +1,1660 @@
+#!/bin/bash
+# Copyright 2020 DSR Corporation
```
ashcherbakov, line 2: Discussed with Aziz:
It would be better to have a more realistic upgrade test which perform update from 0.12 to 1.2 and then 1.2 to 1.3.
As for test readability, we can still have multiple modules (files) for each version, and the master test will call all the modules one by one.
It's OK to fix it in a separate PR.

integration_tests/upgrade/test-upgrade-1.2-to-1.3.sh
Abdulbois: Please look at`integration_tests/upgrade/test-upgrade-1.2-to-1.3.sh:835`

x/pki/handler_test.go
Abdulbois: Please look at `x/pki/handler_test.go:663`, `x/pki/handler_test.go:1513`

integration_tests/run-all.sh
Abdulbois: Please loot at `integration_tests/grpc_rest/pki/helpers.go:754`

x/pki/handler_test.go
```diff
@@ -1462,6 +1510,45 @@ func TestHandler_RevokeX509Cert(t *testing.T) {
...
+		setup.Trustee2.String(), testconstants.PAACertWithSameSubjectID1Subject, testconstants.PAACertWithSameSubjectIDSubjectID, testconstants.Info)
+	_, err = setup.Handler(setup.Ctx, aprRevokeX509Cert)
+	require.NoError(t, err)
+
+	// check that root certificate has been revoked
+	approvedCertificates, _ := queryApprovedCertificates(setup, testconstants.PAACertWithSameSubjectID2Subject, testconstants.PAACertWithSameSubjectIDSubjectID)
+	require.Equal(t, 1, len(approvedCertificates.Certs))
+	require.Equal(t, testconstants.PAACertWithSameSubjectID2Subject, approvedCertificates.Certs[0].Subject)
+	require.Equal(t, testconstants.PAACertWithSameSubjectIDSubjectID, approvedCertificates.SubjectKeyId)
+
```
ashcherbakov, line 1545: I believe we need to query by SubjectKeyId here and make sure that one certificate is returned.

ashcherbakov (approved):

PR #539, opened 2024-02-03
Update the NOC root certificate design in accordance with the latest feedback

akarabashov:
- Added `serial_number`, `info`, and `time` optional fields to the `REVOKE_NOC_X509_ROOT_CERTIFICATE` transaction.
- Added `serial_number` optional field to the `REMOVE_NOC_X509_ROOT_CERTIFICATE` transaction.
- Updated the `REMOVE_NOC_X509_ROOT_CERTIFICATE` description to clarify that deletion also removes revoked certificates.
- Added the `GET_ALL_REVOKED_NOC_X509_ROOT_CERTS` transaction.

### DIFF
--- a/docs/design/noc-root-cert-design.md
+++ b/docs/design/noc-root-cert-design.md
@@ -52,14 +52,20 @@ Revoked NOC root certificates can be re-added using the `ADD_NOC_X509_ROOT_CERTI
 - Parameters:
   - subject: `string` - Base64 encoded subject DER sequence bytes of the certificate.
   - subject_key_id: `string` - Certificate's `Subject Key Id` in hex string format, e.g., `5A:88:0E:6C:36:53:D0:7F:B0:89:71:A3:F4:73:79:09:30:E6:2B:DB`.
+  - serial_number: `optional(string)` - Certificate's serial number.
+  - info: `optional(string)` - Information/notes for the revocation
+  - time: `optional(int64)` - Revocation time (number of nanoseconds elapsed since January 1, 1970 UTC). CLI uses the current time for that field.
 - In State:
   - `pki/RevokedCertificates/value/<subject>/<subject_key_id>`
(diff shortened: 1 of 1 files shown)


docs/design/noc-root-cert-design.md
```diff
@@ -52,24 +52,31 @@ Revoked NOC root certificates can be re-added using the `ADD_NOC_X509_ROOT_CERTI
...
+**Node:** Revoked certificates that match the specified parameters will also be removed.
+
 - Who can send: Vendor account
   - Vid field associated with the corresponding NOC root certificate on the ledger must be equal to the Vendor account's VID.
 - Validation:
   - A NOC root certificate with the provided `subject` and `subject_key_id` must exist in the ledger.
 - Parameters:
   - subject: `string` - Base64 encoded subject DER sequence bytes of the certificate.
   - subject_key_id: `string` - Certificate's `Subject Key Id` in hex string format, e.g., `5A:88:0E:6C:36:53:D0:7F:B0:89:71:A3:F4:73:79:09:30:E6:2B:DB`.
+  - serial_number: `optional(string)` - Certificate's serial number.
```
ashcherbakov, line 77: Let's add `info` field (optional) to be able to describe the reason for remove.

ashcherbakov (commented):
Please also add/fix the following:
- ADD_NOC_X509_ROOT_CERTIFICATE: remove `vid` field from the parameters (vid will be automatically taken Vendor Accuont's VID value)
- Remove command: mention, that child certificates are NOT removed
- Revoke command: add a new optional parameter `revokeChild`, false by default. If true, then all children certs (intermediate, lead) are revoked as well. If false - only the current root cert is revoked.
- Revoke command: specify what happens if serialNumber is not provided (all certs for Subject+SKID are revoked)

docs/design/noc-root-cert-design.md
akarabashov: Done

ashcherbakov (approved):

PR #540, opened 2024-02-08
Support for forward and backward compatibility in DCL schemes

akarabashov:


docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+Schema changes can cover a wide range of modifications with varying impacts on application compatibility and data integrity. Below are use cases with strategies to manage schema changes and ensure compatibility.
+
+## Strategy for Compatible Changes
+
+For changes that are backward-compatible, such as adding optional fields or extending enumerations:
+
+**Strategy steps:**
+
+- One time actions:
+  - Add a version field to DCL schema to track the schema version.
```
ashcherbakov, line 12:
```suggestion
  - Add an optional version field to all DCL schema to track the schema version.
```

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+- For each update:
+  - Update the schema.
+  - Probably need to update transactions and queries.
```
ashcherbakov, line 15: Let's also add the following:
- DCL doesn't fulfill the Schema version automatically
- It will be up to the transaction submitter (Vendor) to specify a correct Schema version
- If Schema Version is not set - then the initial version (version 0 or 1) is assumed
- It will be up to the client application to process the Schema version

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+## Strategy for Compatible Changes
+
+For changes that are backward-compatible, such as adding optional fields or extending enumerations:
+
+**Strategy steps:**
+
+- One time actions:
+  - Add a version field to DCL schema to track the schema version.
+- For each update:
+  - Update the schema.
```
ashcherbakov, line 14:
```suggestion
  - Update the schema by introducing compatible changes (such as adding a new optional field)
```

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+This strategy is straightforward and quick to implement, but only suitable for compatible changes.
+
+## Strategy for Convertible Changes
+
+For changes that affect compatibility but can be converted, like field type changes, renaming fields, and changing enumerations:
+
+**Strategy steps:**
+
+- For each update:
+  - Create a new schema version and state
```
ashcherbakov, line 26:
```suggestion
  - Create a new version of a Schema and state (a new .proto file)
```

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+  - Migrate older states to newer schema version.
```
ashcherbakov, line 27: Probably we need a clarification here.
Does it mean that a data will be represented in two copied: old format AND new format? So it can be queried in either old (legacy) or new way?

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+  - Implement transactions and queries capable of converting data between the latest and older schema version.
```
ashcherbakov, line 28: If we do the migration of old data to the new format, then probably no need to do migration "on the fly" in handlers?
Also, please note that migration on the fly won't allow to use a Light Client feature (as the data is not in the State and proofs can be generated).

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+
+**Strategy steps:**
+
+- One time actions:
+  - Create a more flexible, generic schema structure to hold a wide range of data formats (Can be used [Any](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/any.proto) as described in [ADR-19](https://docs.cosmos.network/v0.47/build/architecture/adr-019-protobuf-state-encoding#usage-of-any-to-encode-interfaces))
+  - Migrate old states to the newer, generic schema.
+  - Implement transactions and queries that can handle data according to its version, including mechanisms for converting generic values into the corresponding schema version.
+- For each update:
+  - Create a new schema version
+  - Probably need to update transactions and queries
```
ashcherbakov, line 44: Please mention cons here, such as that there will be no strict DCL schema anymore.
Also, this change is a breaking change on its own (for all existing clients).
I would keep this option on the doc, but would move it to "Not Recommended for Production" section.

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+While offering a robust solution for handling radical changes, this method requires careful planning and development, which can potentially take a significant amount of time.
+
+## Strategy for Structural Changes
+
+For major organizational changes to the schema, such as splitting or merging schemas:
+
+**Strategy steps:**
+
+- For each update:
+  - Create a new schema version and state
```
ashcherbakov, line 55:
```suggestion
  - Create a new schema version and state (new .proto files)
```

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+  - Migrate older states to newer schema version.
```
ashcherbakov, line 56: Do we really need migration here? It seems we can not always do a migration if for example a new mandatory field is added.

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+  - Implement transactions and queries specifically designed for the reorganized schema.
```
ashcherbakov, line 57: Please clarify, that there will be separate calls for each version, for example
- models/vid/pid
- modelsV2/vid/pid
- modelsV3/vid/pid

As the data is not convertible and not migrated, every entity will be associated with one version only. So, it will be returned by either `models/vid/pid`, `modelsV2/vid/pid` or `modelsV3/vid/pid` depending on what Schema version (.proto version) was used during creation.

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+
+For changes that affect compatibility but can be converted, like field type changes, renaming fields, and changing enumerations:
+
+**Strategy steps:**
+
+- For each update:
+  - Create a new schema version and state
+  - Migrate older states to newer schema version.
+  - Implement transactions and queries capable of converting data between the latest and older schema version.
+
```
ashcherbakov, line 29: Please clarify, that there will be separate calls/requests for each version, for example
- models/vid/pid
- modelsV2/vid/pid
- modelsV3/vid/pid

As the data is convertible and migrated, old (legacy) entities can be returned in either legacy format (for example `models/vid/pid` or `modelsV2/vid/pid`), or a new format (for example `modelsV3/vid/pid`).

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,63 @@
...
+- For each update:
+  - Create a new schema version and state
+  - Migrate older states to newer schema version.
+  - Implement transactions and queries specifically designed for the reorganized schema.
+
+These types of changes are relatively rare. Implementing backward compatibility and migration can be complex.
+
+## Conclusion
+
+To lay the foundation for future compatibility improvements, it's a good idea to start by adding a version field to each schema. This step not only provides compatible updates to schemas, but also prepares the schemas for seamless integration with upcoming compatibility enhancements.
```
ashcherbakov, line 63: Please also add the following:
Non-compatible changes will be handled by creating new Schemas (new .proto files) with separate queries/requests.

docs/design/schema-compatibility.md
akarabashov: Done

docs/design/schema-compatibility.md
akarabashov: Done

docs/design/schema-compatibility.md
akarabashov: Done

docs/design/schema-compatibility.md
akarabashov: Done

docs/design/schema-compatibility.md
akarabashov: Done

docs/design/schema-compatibility.md
akarabashov: Done

docs/design/schema-compatibility.md
akarabashov: Done

docs/design/schema-compatibility.md
```diff
@@ -2,34 +2,54 @@
...
+  - Update older transactions and queries to converting data between the latest and older schema version,  ensuring backward compatibility.
+  - There will be separated API for each version of the schema, for example::
+    - models/vid/pid
+    - modelsV2/vid/pid
+    - modelsV3/vid/pid
 
 The specific implementation details will vary with each change, each time requires an individual approach.
 
-## Strategy for Handling Non-Convertible Changes
+Support for the Light Client feature will not be extended to legacy APIs due to on-the-fly data migration (as the data is not in the State and proofs can be generated).
```
ashcherbakov, line 46: Not sure that it's a good option....

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,119 @@
...
+This strategy focuses on updating the schema without ensuring backward compatibility at the API level. Since the schemas are not compatible, migration is carried out manually through a special transaction.
+
+**Strategy steps:**
+
+- For each update:
+  - Update the schema by introducing changes.
+  - Update transactions and queries if needed.
+  - Add a new transaction to fulfill new required fields (essentially this is a manual migration via transactions)
+
+### Strategy for Non-Compatible changes (Keeping backward compatibility in API)
```
ashcherbakov, line 96: It seems this strategy is suitable for compatible (actually convertable) changes only.
So, probably we can rename the Options here a bit:

- Strategy for Compatible or Convertable changes
  - Option 1: Not keeping backward compatibility in API
  - Option 2: Keeping backward compatibility in API
- Strategy for Non-convertable changes
  - Option 1: Not keeping backward compatibility in API
  - Option 2: Keeping backward compatibility in API

I think we can add the following for the last option:
It's not possible to replace an old version here. "Multiple versions can live in parallel: Strategy for Non-Compatible Changes" options should be used instead.

docs/design/schema-compatibility.md
```diff
@@ -0,0 +1,119 @@
+# Support for forward and backward compatibility in DCL schemes
+
+Schema changes can cover a wide range of modifications with varying impacts on application compatibility and data integrity. Below are use cases with strategies to manage schema changes and ensure compatibility.
+
+## Multiple versions can live in parallel
```
ashcherbakov, line 5: It would be great to add number for every option.
Like

I. Multiple versions can live in parallel
  1. Strategy for Compatible Changes
     - Option A
     - Option B
  2.  Strategy for Non-Compatible Changes
       - Option C
       - Option D

II. New version replaces the legacy one (V2 replaces V1)
   1. Strategy for Compatible or Convertable changes
        - Option E: Not keeping backward compatibility in API
        - Option F: Keeping backward compatibility in API
   2. Strategy for Non-convertable changes
        - Option G: Not keeping backward compatibility in API
        - Option H: Keeping backward compatibility in API

ashcherbakov (approved):

Issue #524, opened 2023-11-10
Root certificates for Network Operational Credential (NOC)


ashcherbakov:
Root CA that NOCs chain up to.
It's similar to ROOT_CERT transactions (see https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/docs/transactions.md#x509-pki), but they are managed (publish/revoke) by Vendors and do not require Trustee approval. 

**Acceptance Criteria:**
- Design the new transaction/requests
  - Most probably new transactions, something like ADD_X509_ROOT_CERT_BY_VENDOR, REVOKE_X509_ROOT_CERT_BY_VENDOR
  - Think if these certificates should be the same as other Root Certs, or separate entities. Probably separate entities, because GET_X509_ROOT_CERT should probably return only Trustee-approved certificates (so called PAAs), not this new ones.
  - The transaction can be sent by Vendor account only
- Generate the new transactions/requests
  - See https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/README-DEV.md#how-to-add-a-new-module-or-transaction
- Implement business logic (handlers)
  - Check that the new PKI cert is a VID-scoped certificate (see https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/x/pki/x509/x509.go#L104) with the same VID as the Account 
- Add Unit Tests. Consider improving/simplifying existing unit tests if it makes sense. Create tests for both positive and negative scenarios.
- Add Integration tests (CLI and REST). Positive and basic negative tests (to check that errors are returned) are enough.
- Update docs
  - https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/docs/transactions.md
  - https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/docs/how-to.md

ashcherbakov:
PR with Design: https://github.com/zigbee-alliance/distributed-compliance-ledger/pull/529
There are open questions: https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/docs/design/noc-root-cert-design.md#questions that needs to be discussed on the DCL TT call.

ashcherbakov:
**Discussed the open questions on DCL TT call:**
- Should the vendor add a revocation distribution point for NOC certificates: **NO**, just revocation/remove (the same as for PAA Roots) is enough
- Should the following queries return NOC Certificate?
  - GET_ALL_SUBJECT_X509_CERTS: **YES**
[...]

PR #550, opened 2024-03-13
#524 Enable revocation of NOC root certificates

Abdulbois:
Enable revoking NOC X509 Root certificates


integration_tests/cli/pki-noc-certs.sh
```diff
@@ -7,6 +7,9 @@ noc_root_cert_1_subject_key_id="44:EB:4C:62:6B:25:48:CD:A2:B3:1C:87:41:5A:08:E7:
 noc_root_cert_1_serial_number="47211865327720222621302679792296833381734533449"
 noc_root_cert_1_subject_as_text="CN=NOC-1,OU=Testing Division,O=Example Company,L=Tashkent,ST=Some State,C=UZ"
 
+noc_root_cert_1_copy_path="integration_tests/constants/noc_root_cert_1_copy"
```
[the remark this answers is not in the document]
akarabashov: After discussions, it was decided to do it in the scope of another task since we have the same certificates types but with a different naming format, it will be necessary to convert the names into a common format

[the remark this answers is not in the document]
akarabashov: > I think that using 'copy' as a postfix might not be the best choice, maybe it would be better to add a part of the serial number as a postfix instead?

After discussions, it was decided to do it in the scope of another task since we have the same certificates types but with a different naming format, it will be necessary to convert the names into a common format

akarabashov (approved):

x/pki/keeper/noc_root_certificates.go
```diff
@@ -70,6 +70,54 @@ func (k Keeper) RemoveNocRootCertificates(
 	))
 }
 
+func (k Keeper) RemoveNocRootCertificate(ctx sdk.Context, vid int32, subject, subjectKeyID string) {
+	certs, found := k.GetNocRootCertificates(ctx, vid)
+	if !found {
+		return
+	}
+
+	certsBefore := len(certs.Certs)
```
akarabashov, line 79: numCertsBefore

x/pki/keeper/noc_root_certificates.go
```diff
@@ -70,6 +70,54 @@ func (k Keeper) RemoveNocRootCertificates(
...
+	for i := 0; i < len(certs.Certs); {
+		cert := certs.Certs[i]
+		if cert.Subject == subject && cert.SubjectKeyId == subjectKeyID {
+			certs.Certs = append(certs.Certs[:i], certs.Certs[i+1:]...)
+		} else {
+			i++
+		}
+	}
```
akarabashov, line 87: Since the filter logic is used in three places, it might be better to implement a generic filter function for it

x/pki/keeper/noc_root_certificates.go
```diff
@@ -70,6 +70,54 @@ func (k Keeper) RemoveNocRootCertificates(
...
+
+	if len(certs.Certs) == 0 {
+		k.RemoveNocRootCertificates(ctx, vid)
+	} else if certsBefore > len(certs.Certs) { // Update state only if any certificate is removed
```
akarabashov, line 91: Should we add a similar check to the `RemoveNocCertificate` function?

Abdulbois: Will be implemented at the next PR of `Revocation of NOC non-root certificates`

PR #571, opened 2024-05-02
Improve error message texts

Abdulbois:


x/pki/keeper/msg_server_remove_noc_x_509_root_cert.go
```diff
@@ -55,18 +71,18 @@ func (k msgServer) RemoveNocX509RootCert(goCtx context.Context, msg *types.MsgRe
 			k.removeApprovedX509Cert(ctx, certID, &aprCerts, msg.SerialNumber)
```
akarabashov, line 71: Can we use nocCerts variable instead of aprCerts?

x/pki/keeper/msg_server_remove_noc_x_509_ica_cert.go
```diff
@@ -64,14 +75,14 @@ func (k msgServer) RemoveNocX509IcaCert(goCtx context.Context, msg *types.MsgRem
 			k.removeApprovedX509Cert(ctx, certID, &aprCerts, msg.SerialNumber)
```
akarabashov, line 75: Can we use icaCerts variable instead of aprCerts?

akarabashov (approved):

Issue #570, opened 2024-05-01
Validate "OtaChecksum" is base64 encoded when adding a Model-Version entry


jcps07:
From conversation in [PR #33256](https://github.com/project-chip/connectedhomeip/pull/33256): 

> [May 1, 2024](https://github.com/project-chip/connectedhomeip/pull/33256#discussion_r1586439969)
> 
> We've seen a few cases where a software update is posted in the DCL and the OtaChecksum is not base64 encoded. What do you think about adding a test that verifies at least that all the software versions for this device model have a correctly formatted checksum? Could potentially also verify other fields as well but this one in particular feels especially important.

I'd be good to add this validation at dcld stage, as it's a requirement by the Matter Spec:

> OtaChecksum
> OtaChecksum SHALL contain the digest of the entire contents of the associated OTA Software
> Update Image under the OtaUrl field, encoded in base64 string representation. The digest SHALL
> have been computed using the algorithm specified in OtaChecksumType. This field SHALL be provided
> if the OtaUrl field is populated.

akarabashov:
PR: https://github.com/zigbee-alliance/distributed-compliance-ledger/issues/570

- Added validation to ensure that 'OtaChecksum' field is base64 encoded data
- Unit and integration tests are added
- Docs are updated

Issue #566, opened 2024-04-23
SchemaVersion of DAC PKI Revocation Distribution Point schema did not get bumped after #557


tcarmelveilleux:
PR #557 added a field to the CRL distribution point schema (`CRLSignerDelegator`) but the SchemaVersion was not bumped. https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9126 will bump the version, but it has to be bumped also in the DCL implementation.

akarabashov:
PR: https://github.com/zigbee-alliance/distributed-compliance-ledger/pull/572

The default version of SchemaVersion for PkiRevocationDistributionPoint is set to 1

PR #572, opened 2024-05-22
improvements #566 #570

Abdulbois:
- Bump schemaVersion of PKI Distribution Point model
- Add base64 format validation for OTAChecksum of Model Version model


x/model/types/errors.go
```diff
@@ -131,3 +132,8 @@ func NewErrModelVersionDeletionCertified(vid interface{}, pid interface{}, softw
 		"Model version associated with vid=%v, pid=%v and softwareVersion=%v has a compliance record and can't be deleted",
 		vid, pid, softwareVersion)
 }
+
+func NewErrOtaChecksumIsNotValid(otaChecksum string) error {
+	return errors.Wrapf(ErrOtaChecksumIsNotValid,
+		"OtaChecksum %v is not base64 encoded", otaChecksum)
```
akarabashov, line 138: I think the name of the function and the error message do not match, we need to correct the message or the name of the function

x/model/keeper/msg_server_model_version.go
```diff
@@ -44,6 +45,11 @@ func (k msgServer) CreateModelVersion(goCtx context.Context, msg *types.MsgCreat
 		return nil, types.NewErrModelVersionAlreadyExists(msg.Vid, msg.Pid, msg.SoftwareVersion)
 	}
 
+	_, err = base64.StdEncoding.DecodeString(msg.OtaChecksum)
+	if err != nil {
+		return nil, types.NewErrOtaChecksumIsNotValid(msg.OtaChecksum)
+	}
```
akarabashov, line 51: Can we move this validation to the 'ValidateBasic()' method?

We also need to validate it when updating the model.

Issue #547, opened 2024-03-06
New fields for Enhanced Setup Flow


koosc:
New fields will need to be added to support the Enhanced Setup Flow. These will contain fields in DeviceModel to identify the options to use for the flow and a new Schema to support storage of the texts that will be used for this flow.

PR in progress: https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/8944

DCL Changes: https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/8944/files#diff-340658b1e5339d59a1bc5b917afe73349e7ef74f0d5e79f3475e5474c35c520a

akarabashov:
PR: https://github.com/zigbee-alliance/distributed-compliance-ledger/pull/577
- Functionality implemented as described in the specification
- Unit and integration tests are added
- Docs are updated.

PR #577, opened 2024-06-03
#547 Add new fields into Model entity

Abdulbois:


x/model/keeper/msg_server_model.go
```diff
@@ -64,6 +65,14 @@ func (k msgServer) CreateModel(goCtx context.Context, msg *types.MsgCreateModel)
 		model.CommissioningModeInitialStepsHint = 1
 	}
 
+	if model.EnhancedSetupFlowOptions == 0 {
+		model.EnhancedSetupFlowTCUrl = msg.EnhancedSetupFlowTCUrl
+		model.EnhancedSetupFlowTCRevision = msg.EnhancedSetupFlowTCRevision
+		model.EnhancedSetupFlowTCDigest = msg.EnhancedSetupFlowTCDigest
+		model.EnhancedSetupFlowTCFileSize = msg.EnhancedSetupFlowTCFileSize
```
akarabashov, line 72: The `EnhancedSetupFlowTCFileSize` field can only be set if `EnhancedSetupFlowOptions` is present.

x/model/keeper/msg_server_model.go
```diff
@@ -144,6 +153,15 @@ func (k msgServer) UpdateModel(goCtx context.Context, msg *types.MsgUpdateModel)
...
 	}
 
+	model.EnhancedSetupFlowOptions = msg.EnhancedSetupFlowOptions
+	if msg.EnhancedSetupFlowOptions == 0 {
+		model.EnhancedSetupFlowTCUrl = msg.EnhancedSetupFlowTCUrl
+		model.EnhancedSetupFlowTCRevision = msg.EnhancedSetupFlowTCRevision
+		model.EnhancedSetupFlowTCDigest = msg.EnhancedSetupFlowTCDigest
+		model.EnhancedSetupFlowTCFileSize = msg.EnhancedSetupFlowTCFileSize
+		model.MaintenanceUrl = msg.MaintenanceUrl
+	}
```
akarabashov, line 163: Maybe we should reset these fields when EnhancedSetupFlowOptions is set?

PR #588, opened 2024-08-01
Model fields fixes

DenisRybas:
- fixed validations for `EnhancedSetupFlowTCUrl, EnhancedSetupFlowTCRevision, EnhancedSetupFlowTCDigest, MaintenanceURL` fields


x/model/types/messages_model.go
```diff
@@ -106,7 +108,7 @@ func (msg *MsgCreateModel) ValidateBasic() error {
 		return err
 	}
 
-	if msg.EnhancedSetupFlowOptions == 0 {
+	if msg.EnhancedSetupFlowOptions&1 == 1 {
```
ashcherbakov, line 111: Do we have a logic and tests that EnhancedSetupFlow fields must be present IF AND ONLY IF EnhancedSetupFlowOptions&1 == 1 when Creating a Model?

x/model/types/messages_model.go
```diff
@@ -203,7 +205,7 @@ func (msg *MsgUpdateModel) ValidateBasic() error {
...
```
ashcherbakov, line 208: Do we have a logic and tests that EnhancedSetupFlow fields must be present IF AND ONLY IF EnhancedSetupFlowOptions&1 == 1 when Updating a Model?

proto/zigbeealliance/distributedcomplianceledger/model/model.proto
```diff
@@ -13,24 +13,25 @@ message Model {
...
-  int32 lsfRevision = 17;
-  string creator = 18 [(cosmos_proto.scalar) = "cosmos.AddressString"];
-  uint32 schemaVersion = 19;
-  int32 enhancedSetupFlowOptions = 20 [(gogoproto.moretags) = "validate:\"omitempty,gte=0,lte=65535\""];
-  string enhancedSetupFlowTCUrl = 21 [(gogoproto.moretags) = "validate:\"required_if=EnhancedSetupFlowOptions 0,omitempty,url,startsnotwith=http:,max=256\""];
-  int32 enhancedSetupFlowTCRevision = 22 [(gogoproto.moretags) = "validate:\"required_if=EnhancedSetupFlowOptions 0,gte=0,lte=65535\""];
-  string enhancedSetupFlowTCDigest = 23 [(gogoproto.moretags) = "validate:\"required_if=EnhancedSetupFlowOptions 0,omitempty,max=128\""];
-  uint32 enhancedSetupFlowTCFileSize = 24 [(gogoproto.moretags) = "validate:\"required_with=EnhancedSetupFlowTCUrl\""];;
-  string maintenanceUrl = 25 [(gogoproto.moretags) = "validate:\"required_if=EnhancedSetupFlowOptions 0,omitempty,url,startsnotwith=http:,max=256\""];
+  int32 discoveryCapabilitiesBitmask = 7 [(gogoproto.moretags) = "validate:\"gte=0,lte=14\""];
```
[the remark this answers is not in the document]
DenisRybas: Fixed

proto/zigbeealliance/distributedcomplianceledger/model/model.proto
```diff
@@ -13,24 +13,25 @@ message Model {
...
+  int32 commissioningCustomFlow = 8;
```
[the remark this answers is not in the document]
DenisRybas: Fixed

x/model/types/messages_model.go
DenisRybas: Added tests

x/model/types/messages_model.go
DenisRybas: Added tests

proto/zigbeealliance/distributedcomplianceledger/model/tx.proto
```diff
@@ -42,11 +42,11 @@ message MsgCreateModel {
   string lsfUrl = 17 [(gogoproto.moretags) = "validate:\"omitempty,url,startsnotwith=http:,max=256\""];
   uint32 schemaVersion = 18 [(gogoproto.moretags) = "validate:\"gte=0,lte=65535\""];
   int32 enhancedSetupFlowOptions = 19 [(gogoproto.moretags) = "validate:\"omitempty,gte=0,lte=65535\""];
```
ashcherbakov, line 44: Do we need to fix startsnotwith=http: for other fields?

proto/zigbeealliance/distributedcomplianceledger/model/tx.proto
```diff
@@ -69,11 +69,11 @@ message MsgUpdateModel {
   uint32 schemaVersion = 15 [(gogoproto.moretags) = "validate:\"gte=0,lte=65535\""];
   uint32 commissioningModeInitialStepsHint = 16;
   int32 enhancedSetupFlowOptions = 17 [(gogoproto.moretags) = "validate:\"omitempty,gte=0,lte=65535\""];
-  string enhancedSetupFlowTCUrl = 18 [(gogoproto.moretags) = "validate:\"required_if=EnhancedSetupFlowOptions 0,omitempty,url,startsnotwith=http:,max=256\""];
-  int32 enhancedSetupFlowTCRevision = 19 [(gogoproto.moretags) = "validate:\"required_if=EnhancedSetupFlowOptions 0,gte=0,lte=65535\""];
-  string enhancedSetupFlowTCDigest = 20 [(gogoproto.moretags) = "validate:\"required_if=EnhancedSetupFlowOptions 0,omitempty,max=128\""];
+  string enhancedSetupFlowTCUrl = 18 [(gogoproto.moretags) = "validate:\"required_if_bit_0_set=EnhancedSetupFlowOptions,omitempty,url,startswith=https:,max=256\""];
```
ashcherbakov, line 72: I'm not sure required_if_bit_0_set is a right validation for the Update command.
All of these fields are optional.
We may have enhancedSetupFlowOptions not specified (assuming the old value is kept), and SOME of the  enhancedSetupFlow fields specified (to update the corresponding values). It seems the current validation will not allow such case.

x/model/keeper/msg_server_model.go
```diff
@@ -149,7 +149,7 @@ func (k msgServer) UpdateModel(goCtx context.Context, msg *types.MsgUpdateModel)
 	}
 
 	model.EnhancedSetupFlowOptions = msg.EnhancedSetupFlowOptions
-	if msg.EnhancedSetupFlowOptions == 0 {
+	if msg.EnhancedSetupFlowOptions&1 == 1 {
```
ashcherbakov, line 152: I think this logic is not correct. msg.EnhancedSetupFlowOptions may not be present in the Update command, meaning that the old value needs to be preserved. On the other hands, only some of the EndhancedSetupFlow values may be present (needs to be updated)

I think the right logic would be the following: 
- get the new values for all EnhancedSetupFlow fields (old one if not in the msg or updated one if in the msg)
- if the new value of EnhancedSetupFlowOptions&1 == 1, make sure all new values for other corresponding  EnhancedSetupFlow are set
- if the new value of EnhancedSetupFlowOptions&1 != 1, make sure all new values for other corresponding  EnhancedSetupFlow are NOT set

We may think how to simplify that logic...

x/model/keeper/msg_server_model.go
```diff
@@ -149,7 +149,7 @@ func (k msgServer) UpdateModel(goCtx context.Context, msg *types.MsgUpdateModel)
 	}
 
 	model.EnhancedSetupFlowOptions = msg.EnhancedSetupFlowOptions
```
ashcherbakov, line 151: Also the following check needs to be added according to the spec:
Update Model: if EnhancedSetupFlowTCRevision is present, it must be equal to oldValue+1. If the oldValue == 0, then new value can be any. If the newValue  ==0, then no updates (OK). If the newValue == oldValue - error (must be oldValue + 1)

proto/zigbeealliance/distributedcomplianceledger/model/tx.proto
ashcherbakov: OK, actually we can keep the current logic, that in order to update one of ENHANCED fields, ALL fields must be present (or none of them).

x/model/keeper/msg_server_model.go
ashcherbakov: Even if we decide to do https://github.com/zigbee-alliance/distributed-compliance-ledger/pull/588#discussion_r1702084104, I think the logic under `else` here is not correct. If Enhanced* fields are not provided, then we should keep old/current values, not override them to empty ones.

proto/zigbeealliance/distributedcomplianceledger/model/tx.proto
DenisRybas: Yep. Fixed

x/model/keeper/msg_server_model.go
DenisRybas: Added validation

proto/zigbeealliance/distributedcomplianceledger/model/tx.proto
ashcherbakov: Had to revert it, as actually it says "All HTTP based URLs SHALL use the https scheme", not that it shall be HTTPS only.

ashcherbakov (approved):

Issue #597, opened 2024-08-19
Improvement Request: Handle Trailing Newline Differences in Certificate Comparison


jcps07:
The DCL currently stores certificates in PEM format as character strings, including newline characters `\n`. When comparing certificates (i.e., validation of certificates when adding CRL Revocation Point), the presence or absence of a trailing newline can cause two otherwise identical certificates to be treated as different.

Recurrent Issue:

    Certificates with a trailing newline at the end are considered different from those without, even if their content is identical.
    This discrepancy can lead to issues when comparing certificates that are essentially the same but formatted slightly differently.

Suggested Improvement:

It would be beneficial if the DCL could ignore differences caused by trailing newlines when comparing certificates. Specifically, I'd be good implementing logic that:

1. Doesn't account for trailing newlines from certificates before comparison.
2. Compares certificates based on their actual content, disregarding the trailing newline as an insignificant difference. (i.e., look for the content within the -----BEGIN CERTIFICATE----- and -----END CERTIFICATE-----  delimiters)

PR #600, opened 2024-08-30
Certificate comparison improvement

DenisRybas:
Now certificates are compared ignoring whitespace characters as described in the issue: https://github.com/zigbee-alliance/distributed-compliance-ledger/issues/597


integration_tests/grpc_rest/pki/helpers.go
```diff
@@ -1687,6 +1687,21 @@ func Demo(suite *utils.TestSuite) {
 	revDistPoints, _ = GetAllPkiRevocationDistributionPoints(suite)
 	require.Equal(suite.T, 1, len(revDistPoints))
 
+	// Add revocation distribution point for PAA by Vendor with certificate with different whitespaces
+	msgAddPkiRevDistPoints = pkitypes.MsgAddPkiRevocationDistributionPoint{
```
ashcherbakov, line 1691: Can you please add a dedicated unit and integration tests for the problem (adding a Distr Point for the same certificate as PAA but with whitespaces)? So that it's clear from the tests that we do cover such case.

DenisRybas: Done

ashcherbakov (approved):

ashcherbakov (approved):

PR #617, opened 2024-11-13
Added index for all certificates by subject key id

Artemkaaas:


ashcherbakov (approved):
It seems need to update openapi as well

PR #644, opened 2025-06-18
bumps ubuntu to 22.04 in release pipeline

andkononykhin:


.github/workflows/release.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         id: current_version
         run: |
           git fetch --all --tags
-          tag="$(git describe --tags --exact-match)"
+          FIXME tag="$(git describe --tags --exact-match)"
```
Shadowsasha, line 44: why do we need this edit?

andkononykhin: right, I forgot to remove, thanks

Shadowsasha (approved):
LGTM

PR #648, opened 2025-06-23
Tests after upgrade

jignatova-dsr:


Toktar:
Can we divide the script to functions please?
For example: setup_network(), test_read_requests(), test_write_requests(), cleanup()

scripts/tests-after-upgrade/tests-after-upgrade.sh
```diff
@@ -0,0 +1,286 @@
...
+result=$(dcld query vendorinfo all-vendors)
+check_response "$result" "\"vendorID\": $vid"
+check_response "$result" "\"vendorID\": $vid_1"
+
+test_divider
+
+# MODEL
+
+echo "Get Model with VID: $vid PID: $pid_for_vid"
+result=$(dcld query model get-model --vid=$vid --pid=$pid_for_vid)
```
Toktar, line 75:
```suggestion
result=$(dcld query model get-model --vid="$vid" --pid="$pid_for_vid")
```
It's better to use quoits for injection and mistake protection

scripts/tests-after-upgrade/tests-after-upgrade.sh
```diff
@@ -0,0 +1,295 @@
...
+
+    cleanup() {
+      echo "Teardown: delete the added NOC Root certificate"
+      result=$(dcld tx pki remove-noc-x509-root-cert --subject="$subject" --subject-key-id="$subject_key_id" --from "$vendor_account" --yes)
+      result=$(get_txn_result "$result")
+      check_response "$result" "\"code\": 0"
+
+      echo "Teardown: delete keys for a $vendor_account"
+      result=$(dcld keys delete "$vendor_account" --yes)
+    }
```
Toktar, line 285: We can move the definition of `cleanup` method right after adding the keys
And call it right after with `trap cleanup EXIT`
Then, if something go wrong in the test, we can be sure that keys and certificate will be removed

PR #663, opened 2025-07-19
Add specification version to model version schema

Shadowsasha:


integration_tests/cli/model-validation-cases.sh
```diff
@@ -508,3 +511,83 @@ result=$(echo "test1234" | dcld tx model update-model-version --vid=$vid_1 --pid
...
+check_response_and_report "$result" "\"firmwareInformation\": \"123456789012345678901234567890123456789012345678901234567890123\""
+check_response_and_report "$result" "\"softwareVersionValid\": true"
+check_response_and_report "$result" "\"otaUrl\": \"https://ota.url.info\""
+check_response_and_report "$result" "\"otaFileSize\": \"123456789\""
+check_response_and_report "$result" "\"otaChecksum\": \"SGVsbG8gd29ybGQh\""
+check_response_and_report "$result" "\"otaChecksumType\": 1"
+check_response_and_report "$result" "\"releaseNotesUrl\": \"https://release.notes.url.info\""
+check_response_and_report "$result" "\"maxApplicableSoftwareVersion\": 32"
+check_response_and_report "$result" "\"minApplicableSoftwareVersion\": 5"
+check_response_and_report "$result" "\"specificationVersion\": 33"
```
Toktar, line 591: It's absolutely non-critical, but I suggest to use constant values for these cases instead of in-line numbers

Toktar (commented):
Thank you for the great PR! Just small comments about tests above. Could you please also add this parameter to `integration_tests/grpc_rest/model/helpers.go` to `NewMsgCreateModelVersion` with a constant value from `integration_tests/constants/constants.go`

Toktar (approved):

PR #671, opened 2025-08-07
adding test coverage

Shadowsasha:


integration_tests/deploy/test_deploy.sh
```diff
@@ -29,29 +29,29 @@ VN_NAME="test_deploy_vn"
...
-GOBIN=${GOBIN:-${GOPATH}/bin}
 
-mkdir -p "$GOBIN"
+mkdir -p "$DCL_TMP_DIR"
 
 docker build -f integration_tests/deploy/Dockerfile-build -t dcl-deploy-build .
 docker container create --name dcl-deploy-build-inst dcl-deploy-build
-docker cp dcl-deploy-build-inst:/go/bin/dcld "$GOBIN"/
-docker cp dcl-deploy-build-inst:/go/bin/cosmovisor "$GOBIN"/
+docker cp dcl-deploy-build-inst:/go/bin/dcld "$DCL_TMP_DIR"/
```
Toktar, line 53: Could you please explain why we need to change GOBIN path to  `/tmp/dcl_tmp`?

.github/workflows/verify.yml
```diff
@@ -103,9 +104,27 @@ jobs:
...
+        with:
+          config: ./.github/.testcoverage.yml
+      - name: Generate html test coverage
+        shell: bash
+        run: go tool cover -html=cover.out -o coverage.html
+      - name: Upload html coverage
+        uses: actions/upload-artifact@v4
+        with:
+          name: unit_coverage
+          path: coverage.html
```
Toktar, line 122: Do we need .html reports for unit and integration tests separately, or we can just leave one file with merged results?

integration_tests/deploy/test_deploy.sh
Shadowsasha: It turned out that the local version was overwritten by the version from the container, which is mildly not the expected behavior. In fact, any folder is suitable for us, but not `GOBIN` folder, so tmp was selected.

Toktar (approved):

PR #699, opened 2025-10-13
Change default value for CommissioningModeSecondaryStepsHint to 4

Toktar:


x/dclupgrade/types/message_propose_upgrade.go
```diff
@@ -94,7 +94,7 @@ func ValidateBinaries(msg *MsgProposeUpgrade, gitBaseURL string) error {
 
 		urlGitTag := partsURL[7]
 
-		if msg.Plan.Name != urlGitTag {
```
ashcherbakov, line 97: This fix will open a possibility for upgrade to plan `v1.6` with the both version (tag) 1.6.1 and 1.6.2, right? Can it lead to some issues?

Toktar: It can. But at the same time, right now we can't test any dev upgrades, only already released. And it would be good to test a release build before releasing.
But you are right, and it's a tricky question. If you think that it's better to revert this change, I'll do this, no problem

ashcherbakov: Looks like as long as we use major.minor.patch in upgrade names we should be good.

ashcherbakov (approved):

PR #707, opened 2026-01-15
Enhance PKI module validations.

### LINKED ISSUE #702: Improve PKI module validation of certificates

Abdulbois:
- Remove decoding certificate from ValidateBasic methods
- Refactor X.509 certificate handling by replacing `DecodeX509Certificate` with `ParseAndValidateCertificate`
- Make regex methods statically compiled
- Make instantiation of validator only once 
- Refactor Add PKI revocation distribution point package

Closes #702


x/pki/x509/x509.go
```diff
@@ -40,8 +40,11 @@ type Certificate struct {
 }
 
 const (
-	Mvid = "Mvid"
-	Mpid = "Mpid"
+	Mvid             = "Mvid"
+	Mpid             = "Mpid"
+	MaxCertSize      = 100 * 1024 // 100 KB
```
ashcherbakov, line 45: I think it makes sense to double-check max size in the Matter spec (in case it's mentioned there) https://github.com/CHIP-Specifications/connectedhomeip-spec.

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -30,7 +24,19 @@ func (k msgServer) AddPkiRevocationDistributionPoint(goCtx context.Context, msg
...
 
-	// compare VID in message and Vendor acount
+	// decode CrlSignerCertificate
+	crlSignerCertificate, err := x509.ParseAndValidateCertificate(msg.CrlSignerCertificate)
+	if err != nil {
+		return nil, pkitypes.NewErrInvalidCertificate(err)
+	}
+
+	// verify CrlSignerCertificate
+	err = msg.VerifyCrlSignerCertificate(crlSignerCertificate)
```
ashcherbakov: I think it would be better to move `VerifyCrlSignerCertificate` to this file (from types/message_...)

x/pki/types/message_add_pki_revocation_distribution_point_test.go
```diff
@@ -177,22 +177,6 @@ func TestMsgAddPkiRevocationDistributionPoint_ValidateBasic(t *testing.T) {
 			},
 			err: pkitypes.ErrInvalidRevocationType,
 		},
-		{
-			name: "pid not empty when IsPAA is true",
```
ashcherbakov, line 181: Do we have replacement for these tests (on `tests/hander_...` level)?

x/pki/types/message_propose_add_x_509_root_cert_test.go
```diff
@@ -64,17 +63,6 @@ func TestMsgProposeAddX509RootCert_ValidateBasic(t *testing.T) {
 			},
 			err: validator.ErrFieldLowerBoundViolated,
 		},
-		{
-			name: "invalid VID",
```
ashcherbakov, line 68: Do we have replacement for these tests (on tests/hander_... level)?

x/pki/tests/handler_assign_vid_test.go
```diff
@@ -179,7 +179,6 @@ func TestHandler_AssignVid_MessageVidAndCertificateVidNotEqual(t *testing.T) {
 
 	// propose and approve x509 root certificate
 	rootCert := utils.RootDaCertificateWithNumericVid(setup.Trustee1)
-	rootCert.Vid = 0
```
ashcherbakov, line 182: Is it intended that we remove this line?

utils/validator/validator.go
```diff
@@ -26,16 +26,16 @@ import (
...
 var (
-	uni *ut.UniversalTranslator
-	vl  *validator.Validate
+	vl    *validator.Validate
+	trans ut.Translator
 )
 
 //nolint:wrapcheck,errcheck
-func Validate(s interface{}) error {
+func init() {
```
ashcherbakov, line 34: How/where is this `init` method called?

x/pki/x509/x509.go
Abdulbois: Found only below [notes](https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/08294f091a5e78e8dd7b3ec47a9a22c21b2ee1b3/policies/matter_certificate_policy.adoc#71-certificate-profile)
> As required by the Matter specification, all Matter PKI certificates SHALL NOT be longer than 600 bytes when represented as ASN.1 DER encoding.

Probably it will be ~850 bytes in pem format

x/pki/tests/handler_assign_vid_test.go
Abdulbois: Yes, as we pass cert's vid as message vid inside `RootDaCertificateWithNumericVid`

Before, validation about cert.vid == msg.vid was performed in the static layer, which is why the test passed.

utils/validator/validator.go
Abdulbois: Go runtime automatically calls `init` when package is imported

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: Done

x/pki/types/message_add_pki_revocation_distribution_point_test.go
Abdulbois: Moved them to new `message_add_pki_revocation_distribution_point_test.go` file

x/pki/types/message_propose_add_x_509_root_cert_test.go
Abdulbois: Yes, [see](https://github.com/Abdulbois/distributed-compliance-ledger/blob/0e0a5ef7ea790cd6f6fb1528966e810067a30076/x/pki/tests/handler_assign_vid_test.go#L175)

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -144,3 +154,148 @@ func (k msgServer) checkCRLSignerNonRootCert(ctx sdk.Context, crlSignerCertifica
...
+		if ext.Id.Equal(oidKeyUsage) {
+			isCritical = ext.Critical
+
+			break
+		}
+	}
+
+	if !isCritical {
+		return pkitypes.NewErrCRLSignerCertificateInvalidFormat("Basic Constraint extension SHALL be marked critical")
+	}
```
AzizMukhtorjonov: It is said that "Key Usage extension SHOULD be marked critical" but the error says "Basic Constraint extension SHALL be critical".
Is it intentional? Why?
If not - fix the error please

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -144,3 +154,148 @@ func (k msgServer) checkCRLSignerNonRootCert(ctx sdk.Context, crlSignerCertifica
...
+	if vid > 0 && vid != msg.Vid {
+		return pkitypes.NewErrCRLSignerCertificateVidNotEqualMsgVid(vid, msg.Vid)
+	}
+
+	if !cert.IsSelfSigned() {
+		err = VerifyCRLSignerCertFormat(cert)
+		if err != nil {
+			return err
+		}
+	}
```
AzizMukhtorjonov, line 191: Should this check come before or after pids & vids checks? Why?

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -144,3 +154,148 @@ func (k msgServer) checkCRLSignerNonRootCert(ctx sdk.Context, crlSignerCertifica
...
+
+func VerifyCRLSignerCertFormat(certificate *x509.Certificate) error {
+	if certificate.SubjectKeyID == "" {
+		return pkitypes.NewErrWrongSubjectKeyIDFormat()
+	}
+
+	cert := certificate.Certificate
+	if cert.Version != 3 {
+		return pkitypes.NewErrCRLSignerCertificateInvalidFormat(
+			"The version field SHALL be set to 2 to indicate v3 certificate",
```
AzizMukhtorjonov: What about creating new method for error with InvalidVersion instead of InvalidFormat?

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -144,3 +154,148 @@ func (k msgServer) checkCRLSignerNonRootCert(ctx sdk.Context, crlSignerCertifica
...
+		)
+	}
+
+	if cert.SignatureAlgorithm != x509std.ECDSAWithSHA256 {
+		return pkitypes.NewErrCRLSignerCertificateInvalidFormat(
+			"The signature field SHALL contain the identifier for signatureAlgorithm ecdsa-with-SHA256",
```
AzizMukhtorjonov, line 252: What about creating new method for error? Is it really error with invalid format?

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -144,3 +154,148 @@ func (k msgServer) checkCRLSignerNonRootCert(ctx sdk.Context, crlSignerCertifica
...
+	if ecdsaPubKey.Curve != elliptic.P256() {
+		return pkitypes.NewErrCRLSignerCertificateInvalidFormat(
+			"The public key must use prime256v1 curve",
+		)
+	}
+
+	// Basic Constraint extension should be marked critical and have the cA field set to false
+	if !cert.BasicConstraintsValid || cert.IsCA {
+		return pkitypes.NewErrCRLSignerCertificateInvalidFormat(
+			"Basic Constraint extension's cA field SHALL be set to FALSE",
```
AzizMukhtorjonov, line 274:
```suggestion
			"Basic Constraint extension's CA field SHALL be set to FALSE",
```

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -144,3 +154,148 @@ func (k msgServer) checkCRLSignerNonRootCert(ctx sdk.Context, crlSignerCertifica
...
+
+	// Key Usage extension should be marked critical
+	isCritical := false
+	for _, ext := range cert.Extensions {
+		if ext.Id.Equal(oidKeyUsage) {
+			isCritical = ext.Critical
+
+			break
+		}
+	}
```
AzizMukhtorjonov, line 286:
```suggestion
	isCritical := false
	for _, ext := range cert.Extensions {
		if ext.Id.Equal(oidKeyUsage) {
			isCritical = ext.Critical
			break
		}
	}
```

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -144,3 +154,148 @@ func (k msgServer) checkCRLSignerNonRootCert(ctx sdk.Context, crlSignerCertifica
...
+
+	if !isCritical {
+		return pkitypes.NewErrCRLSignerCertificateInvalidFormat("Basic Constraint extension SHALL be marked critical")
+	}
+
+	if cert.KeyUsage&x509std.KeyUsageCRLSign == 0 {
```
AzizMukhtorjonov, line 292:
```suggestion
	if cert.KeyUsage & x509std.KeyUsageCRLSign == 0 {
```

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -144,3 +154,148 @@ func (k msgServer) checkCRLSignerNonRootCert(ctx sdk.Context, crlSignerCertifica
...
+		return pkitypes.NewErrCRLSignerCertificateInvalidFormat("The cRLSign bits SHALL be set in the KeyUsage bitstring")
+	}
+
+	if cert.KeyUsage&^(x509std.KeyUsageCRLSign|x509std.KeyUsageDigitalSignature) != 0 {
```
AzizMukhtorjonov, line 296:
```suggestion
	if cert.KeyUsage &^ (x509std.KeyUsageCRLSign|x509std.KeyUsageDigitalSignature) != 0 {
```

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
```diff
@@ -144,3 +154,148 @@ func (k msgServer) checkCRLSignerNonRootCert(ctx sdk.Context, crlSignerCertifica
...
+		return pkitypes.NewErrCRLSignerCertificateInvalidFormat("The KeyUsage bitstring can only include the cRLSign and digitalSignature bits")
```
AzizMukhtorjonov, line 297:
```suggestion
		return pkitypes.NewErrCRLSignerCertificateInvalidFormat("The KeyUsage bitstring can only include the CRLSign and digitalSignature bits")
```

x/pki/x509/x509.go
```diff
@@ -40,8 +40,11 @@ type Certificate struct {
...
+	MaxSANCount      = 100
+	MaxSubjectFields = 50
```
AzizMukhtorjonov, line 47: Afaik, spec does not specify exact numbers for SAN and subject fields. What are the reasons for these numbers?

x/pki/x509/x509.go
```diff
@@ -196,3 +202,42 @@ func (c Certificate) IsSelfSigned() bool {
...
+	sanCount := len(cert.Certificate.DNSNames) + len(cert.Certificate.EmailAddresses) + len(cert.Certificate.IPAddresses) + len(cert.Certificate.URIs)
+	if sanCount > MaxSANCount {
+		return nil, pkitypes.NewErrInvalidCertificate(fmt.Sprintf("SAN count (%d) exceeds maximum limit of %d", sanCount, MaxSANCount))
+	}
+
+	// 3. Check Subject Fields count
+	subjectFieldCount := len(cert.Certificate.Subject.Names)
+	if subjectFieldCount > MaxSubjectFields {
+		return nil, pkitypes.NewErrInvalidCertificate(fmt.Sprintf("subject field count (%d) exceeds maximum limit of %d", subjectFieldCount, MaxSubjectFields))
+	}
```
AzizMukhtorjonov, line 240:
```
If the subject field
   contains an empty sequence, then the issuing CA MUST include a
   subjectAltName extension that is marked as critical
``` ([ref](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6))

Should we add this check? Or is it abundant?

AzizMukhtorjonov (changes requested):

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: Updated comment

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: The purpose of checking against pid and vid is firstly to verify that the transaction message was sent correctly.

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: Yes, we can, but we also need to create a new error code and type for CRL certificate-related errors. Currently we have only one error type&code for it.

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: Yes, we can, but we also need to create a new error code and type for CRL certificate-related errors. Currently we have only one error type&code for it.

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: it is not typo, cA is correctly declared

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: is not allowed by `gofmt`

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: is not allowed by `gofmt`

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: it is not typo, cRLSign is correctly declared

x/pki/x509/x509.go
Abdulbois: It is a kind of recommend values 
Any certificate with 50+ subject fields is highly suspicious and likely a non-standard payload

x/pki/x509/x509.go
Abdulbois: I don't think this case is relevant here.

AzizMukhtorjonov (approved):

x/pki/types/message_propose_add_x_509_root_cert_test.go
ashcherbakov: But this is for AssignVID command, and we need for ProposeAddX509Root in https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/x/pki/tests/handler_propose_paa_cert_test.go

x/pki/types/message_add_pki_revocation_distribution_point_test.go
ashcherbakov: Can you please clarify it? Do we have such negative tests in `‎x/pki/tests/handler_update_revocation_test.go‎`?

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: Invalid Format means that passed certificate contains non-expected/invalid value

x/pki/keeper/msg_server_add_pki_revocation_distribution_point.go
Abdulbois: Updated

ashcherbakov (approved):

PR #708, opened 2026-01-21
Remove checking Model Version while adding Device Software Compliance record

### LINKED ISSUE #706: Remove Model/Model-Version dependency for adding DeviceSoftwareCompliance records
Currently, a vendor must first create entries in the DeviceModel and DeviceSoftwareVersionModel schemas before a Certification Center can add the corresponding DeviceSoftwareCompliance record. This creates a dependency that is enforced by the DCL, not by the Matter specification itself.

The Matter specification does not require a DeviceSoftwareVersionModel to exist prior to certification. It only defines a DeviceSoftwareCompliance record as being uniquely identified by the combination of VID, PID, and Software Version. As a result, the current DCL behavior introduces an additional prerequisite that is not mandated by the spec.

DCL-TT should review options that would allow the Certification Center account operated by the Alliance to add DeviceSoftwareCompliance records for certified products without requiring the prior existence of DeviceModel and DeviceSoftwareVersionModel entries. One possible approach would be to allow the Certification Center to create the associated Model and Model-Version records with minimal required information when they do not already exist. Another option would be to allow the Certification Center to add DeviceSoftwareCompliance records even when the corresponding Model and Model-Version records are missing.

Abdulbois:
- Enable adding Device Software Compliance Info even Model Version does not exist on the ledger
- Add extra validation  while Adding Model Version - when Compliance Info already exists, check that its software version string and cd version number matches
- Update tests and docs

Closes #706

### DIFF
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -307,8 +307,6 @@ dcld tx vendorinfo update-vendor --vid=<uint16> ... --from=<vendor-admin-account
 
 This step is needed for on-ledger certification use case only, see [use_cases_device_on_ledger_certification](use_cases/use_cases_device_on_ledger_certification.png).
 
-The corresponding model and the version must be present on the ledger.
-
 ```bash
 dcld tx compliance certify-model --vid=<uint16> --pid=<uint16> --softwareVersion=<uint32> --softwareVersionString=<string>  --certificationType=<matter|zigbee> --certificationDate=<rfc3339 encoded date> --reason=<string> --from=<account>
 ```
--- a/docs/transactions/compliance.md
+++ b/docs/transactions/compliance.md
@@ -11,8 +11,6 @@ Attests compliance of the Model Version to the ZB or Matter standard.
 `REVOKE_MODEL_CERTIFICATION` should be used for revoking (disabling) the compliance.
 It's possible to call `CERTIFY_MODEL` for revoked model versions to enable them back.
 
-The corresponding Model and Model Version must be present on the ledger.
-
 It must be called for every compliant device for use cases where compliance
 is tracked on ledger.
 
--- a/docs/transactions/compliance.md
+++ b/docs/transactions/compliance.md
@@ -11,6 +11,8 @@ Attests compliance of the Model Version to the ZB or Matter standard.
 `REVOKE_MODEL_CERTIFICATION` should be used for revoking (disabling) the compliance.
 It's possible to call `CERTIFY_MODEL` for revoked model versions to enable them back.
 
+The corresponding Model and Model Version are not required to be present in the ledger. It can be added later by Vendors.
+
 It must be called for every compliant device for use cases where compliance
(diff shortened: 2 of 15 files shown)


docs/how-to.md
```diff
@@ -307,8 +307,6 @@ dcld tx vendorinfo update-vendor --vid=<uint16> ... --from=<vendor-admin-account
```
ashcherbakov, line 307: Please also update https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/docs/transactions/compliance.md#CERTIFY_MODEL (transaction.md is the main source of documentation, so please make sure it's always up to date).

ashcherbakov (changes requested):

docs/how-to.md
Abdulbois: Updated

ashcherbakov (approved):

AzizMukhtorjonov (approved):

docs/transactions/compliance.md
```diff
@@ -11,8 +11,6 @@ Attests compliance of the Model Version to the ZB or Matter standard.
 `REVOKE_MODEL_CERTIFICATION` should be used for revoking (disabling) the compliance.
 It's possible to call `CERTIFY_MODEL` for revoked model versions to enable them back.
```
ashcherbakov, line 13: Please also extend model docs to mention the logic for SW Version String and CDVersionNumber validation (that they must match corresponding compliance record if it's present).

docs/transactions/compliance.md
```diff
@@ -11,8 +11,6 @@ Attests compliance of the Model Version to the ZB or Matter standard.
...
 
-The corresponding Model and Model Version must be present on the ledger.
-
```
ashcherbakov, line 15: Perhaps makes sense to explicitly mention that corresponding Model and ModelVersion don't need to be present on DCL and can be added later by the Vendor.

docs/transactions/compliance.md
Abdulbois: updated

docs/transactions/compliance.md
Abdulbois: updated

x/compliance/client/cli/tx_certify_model.go
```diff
@@ -41,7 +41,7 @@ func CmdCertifyModel() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "certify-model",
-		Short: "Certify an existing model. Note that either corresponding model version and test results or revocation info must be present on ledger",
+		Short: "Certify an existing model",
```
jcps07, line 44:
```suggestion
		Short: "Certify a model-version",
```

Abdulbois: Updated in #713

PR #714, opened 2026-03-30
Support for Ubuntu 24.04 and Infrastructure/Config Improvements

### LINKED ISSUE #705: Ubuntu 24.04 and corresponding IaC and config improvements
Need to support latest LTS Ubuntu(24.04 instead of 20.04) for 

- Build and Docker (such as https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/Dockerfile)
- CI/CD pipelines https://github.com/zigbee-alliance/distributed-compliance-ledger/tree/master/.github
  - Make sure we support both 20.04 and 24.04, and verify both
- Local development https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/README-DEV.md
- IaC scripts  https://github.com/zigbee-alliance/distributed-compliance-ledger/tree/master/deployment 
  - Ansible
  - Terraform for AWS, Azure, GCP  
- Assist community (node admins) in update of all nodes
- Make sure Ubuntu 20.04 is still working, as there can be nodes running on 20.04 

Other things to consider in release/CD:
- archiving the cosmovisor and the dcld together, instead of one dcld to  .ubuntu.tar.gz and .macos.tar.gz
- would be nice to specify the ubuntu version in the  .ubuntu.tar.gz

Abdulbois:
* Ubuntu 24.04 Migration:
       * Updated Dockerfile and Dockerfile-build to support Ubuntu 24.04 via build arguments.
       * Updated CI workflows (release.yml, verify.yml) to use Ubuntu 24.04 as the base for builds and testing.
       * Enabled releasing artifacts for both(20.04 and 24.04) Ubuntu versions.
       * Add doc about upgrading Ubuntu on active/running nodes
   * Infrastructure as Code (IaC):
       * Make Ubuntu 24.04 the base image for the cloud instances.
   * Build System & Localnet:
       * Modified Makefile to set `CGO_ENABLED=0` by default, improving binary portability (except when cleveldb is used).
       * Improved `docker-compose.yml` to allow running a mixed-OS cluster (20.04 and 24.04) for compatibility testing.
       * Enhanced `Makefile.test` and `localnet` scripts (`genlocalnetconfig.sh`, `genlocalnetappbins.sh`) for better robustness and OS version handling.
   * Integration Testing & Upgrades:
       * Added a new end-to-end upgrade test from version 1.5.2 to 1.6.0 (integration_tests/upgrade/09-test-upgrade-1.5.2-to-1.6.0.sh).
       * Refactored and renumbered existing upgrade tests.
       * Updated upgrade scripts to support the new versioning and Ubuntu 24.04 binaries.

Closes #705


deployment/ansible/inventory/group_vars/all.yml
```diff
@@ -53,6 +55,8 @@ dcld_checksums:
   0.7.0: 50708d4f7e00da347d4e678bf26780cd424232461c4bb414f72391c75e39545a
   0.6.0: 1ab6b9084fc0444858307f13b708d9f275d548dd2cfb3bc7d8e10d59897d7278
 cosmovisor_checksums:
+  1.5.2: 9d62e0a2e37b9ccc5ec3aa7220ca04136e47ef015281d88516605738a86de93f
+  1.5.1: 9d62e0a2e37b9ccc5ec3aa7220ca04136e47ef015281d88516605738a86de93f
```
AzizMukhtorjonov, line 59: Are they really the same?

deployment/terraform/gcp/terraform.tfvars
```diff
@@ -1,4 +1,4 @@
-region_1 = "us-east2"
+region_1 = "us-east1"
```
AzizMukhtorjonov, line 1: Why change region?

docs/upgrading-ubuntu-on-running-nodes.md
```diff
@@ -0,0 +1,97 @@
...
+
+    Bellow are the steps for AWS Lightsail.
+
+    1.1 Open `AWS Lightsail` console and select running instance.
+
+    1.2 From bottom navigation bar, select `Snapshots` tab.
+
+    1.3 Click `Create snapshot` and give a name then click to `Create`.
+
+    1.4 Wait until Snapshotting finishes.
```
AzizMukhtorjonov: Is it really necessary to include this step? Looks redundant, but up to you, I have no concerns about keeping or removing

docs/upgrading-ubuntu-on-running-nodes.md
AzizMukhtorjonov, line 1: Please make a directory with docs on upgrade and split docs based on platform (AWS, AWS Lightsail, Azure, GCP)

docs/upgrading-ubuntu-on-running-nodes.md
```diff
@@ -0,0 +1,97 @@
...
+
+2. **Connect to running cloud instance**. Connect using SSH. Below are the steps for AWS Lightsail.
```
AzizMukhtorjonov: Please rephrase "Connect using SSH". Connect via SSH is not necessary, only required item is to have an access to cli on server

docs/upgrading-ubuntu-on-running-nodes.md
```diff
@@ -0,0 +1,97 @@
...
+    ```bash
+    sudo lsb_release -a
+    ```
+    
+    5.2 Verify that the node is running.
+    ```bash
+    dcld status
+    ```
+    
+6. **Upgrade to 24.04**. To upgrade to next `24.04` release, please repeat above steps, you can skip `4.5` step.
```
AzizMukhtorjonov: Please clarify points of start and end.
Is it necessary to take snapshot after upgrade? Can it be skipped and fail case will require start from scratch?

integration_tests/prepare-dcld-versions.sh
```diff
@@ -24,13 +24,22 @@ log() {
 
 mkdir -p $TMP_DCLD_BINS_DIR
 
-for version in 0.12.0 0.12.1 1.2.2 1.4.3 1.4.4 1.5.1 1.5.2; do
+for version in 0.12.0 0.12.1 1.2.2 1.4.3 1.4.4 1.5.1 1.5.2 1.6.0-0.dev.1; do
```
AzizMukhtorjonov: Should it be updated to 1.6.0 before release?

docker-compose.yml
```diff
@@ -15,8 +15,11 @@
 
 services:
   node0:
-    image: dcledger
-    build: .
+    image: dcledger-20.04
+    build:
+      context: .
+      args:
+        UBUNTU_VERSION: 20.04
```
AzizMukhtorjonov, line 22: Why 20?

docker-compose.yml
```diff
@@ -31,8 +34,11 @@ services:
     command: /var/lib/dcl/./node_helper.sh
 
   node1:
-    image: dcledger
-    build: .
+    image: dcledger-20.04
+    build:
+      context: .
+      args:
+        UBUNTU_VERSION: 20.04
```
AzizMukhtorjonov: Why 20?

README.md
```diff
@@ -7,7 +7,7 @@
 If you are interested in how to build and run the project locally, please look at [README-DEV](README-DEV.md)
 
 Please note, that the only officially supported platform now is Linux.
-It's recommended to develop and deploy the App on Ubuntu 18.04 or Ubuntu 20.04.
+It's recommended to develop and deploy the App on Ubuntu 20.04 or Ubuntu 24.04.
```
AzizMukhtorjonov: Remove recommended Ubuntu 20 and rather add it as supported. Ubuntu 20 has lost its support and is not recommended

AzizMukhtorjonov (changes requested):

deployment/ansible/inventory/group_vars/all.yml
Abdulbois: Yes

deployment/terraform/gcp/terraform.tfvars
Abdulbois: Seems `us-east2` does not exist anymore. See https://docs.cloud.google.com/compute/docs/regions-zones#available

Also, I had kind of 'not found' issue while testing trf scripts

docs/upgrading-ubuntu-on-running-nodes.md
Abdulbois: Added separate instructions for 1, 2 and 4.4 steps for each provider(AWS, GCP and Azure)

docker-compose.yml
Abdulbois: To show the compatibility of nodes with different OS versions

docker-compose.yml
Abdulbois: To show the compatibility of nodes with different OS versions

README.md
Abdulbois: Updated

integration_tests/prepare-dcld-versions.sh
Abdulbois: Yes, left `TODO` about that

docs/upgrading-ubuntu-on-running-nodes.md
Abdulbois: Updated

docs/upgrading-ubuntu-on-running-nodes.md
Abdulbois: Updated

docs/upgrading-ubuntu-on-running-nodes.md
```diff
@@ -0,0 +1,103 @@
+# Node OS Upgrade: Upgrading DCL Cloud Instance from Ubuntu 20.04 to Ubuntu 24.04
+
+This document outlines the procedure for upgrading an active Distributed Compliance Ledger (DCL) node from Ubuntu 20.04 to Ubuntu 24.04.
+
```
ashcherbakov, line 4: It looks like the instruction is specific to AWS Lightsail (which is probably no the default way for most of the nodes). 
How the instruction will look like for, for example, EC2?

docs/upgrading-ubuntu-on-running-nodes.md
```diff
@@ -0,0 +1,103 @@
...
+    3.3 After the package upgrade, reboot the instance:
+    ```bash
+    reboot
+    ```
+
+4.  **First Release Upgrade (20.04 to 22.04)**: Initiate the upgrade to Ubuntu 22.04.
+    
+    4.1 Connect to the instance using SSH:
+    ```bash
+    ssh -i "private-key.pem" ubuntu@{{ip or hostname}}
```
ashcherbakov: What if the username is not `ubuntu`?

Abdulbois: Updated

docs/upgrading-ubuntu-on-running-nodes.md
Abdulbois: Added instructions for each provider(AWS, GCP and Azure) for 1, 2 and 4.4 steps

deployment/terraform/gcp/public-sentries/main.tf
```diff
@@ -39,7 +39,7 @@ locals {
 data "google_compute_image" "ubuntu" {
   most_recent = true
   project     = "ubuntu-os-cloud"
-  filter      = "(family = \"${var.os_family}\") AND (architecture = \"X86_64\")"
+  family      = var.os_family
```
andkononykhin: why did you remove architecture constraint here?

deployment/scripts/install.sh
```diff
@@ -69,12 +69,12 @@ function main {
...
             # shellcheck disable=SC1091
             source /etc/os-release || source /usr/lib/os-release
-            distro="${ID,,}"
+            distro="${ID,,}-${VERSION_ID}"
         fi
 
         info "Downloading the DCL binary for $distro..."
         mkdir -p "$dest"
-        url="https://github.com/zigbee-alliance/distributed-compliance-ledger/releases/download/v$version/dcld.$distro.tar.gz"
+        url="https://github.com/zigbee-alliance/distributed-compliance-ledger/releases/download/v$version/$distro.tar.gz"
```
andkononykhin, line 77: why `dcld` prefix is removed?

deployment/ansible/inventory/group_vars/all.yml
andkononykhin: do I remember that correctly: it is not a version of cosmovisor but a dcld release instead where we attach cosmovisor as well?

deployment/ansible/inventory/group_vars/all.yml
```diff
@@ -23,9 +23,9 @@ accounts:
 
 chain_id: testnet-2.0
 company_name: CSA
-type_name: "" # "ON", "Priv-SN", "Pub-SN", "SeedN" or "VN"
-dcld_version: 1.4.4
-cosmovisor_version: 1.5.0-0.dev.3
+type_name: "ON" # "ON", "Priv-SN", "Pub-SN", "SeedN" or "VN"
```
andkononykhin: does it mean we want to set "observer" as a default value?

andkononykhin (commented):
@Abdulbois Hello. Looks good. But I've posted few questions. Also there is more major one:
- looks like the changes covers shift to ubuntu 24.04 for build, test and deploy for new deployments. What about existent environments: is it considered as a separate change / docs update ?

deployment/scripts/install.sh
Abdulbois: I have updated publishing release artifacts for each OS(ubuntu 20.04/24.04, macOS), where all dcld and cosmovisor artifacts are archived together within one archive

deployment/terraform/gcp/public-sentries/main.tf
Abdulbois: Because of GCP image namings for Ubuntu 24.04, `X86_64` is no longer valid, but instead `amd64` or `arm` arch are used. That's why `architecture` was removed

Do you have another approach/solution to declare it better?

deployment/ansible/inventory/group_vars/all.yml
Abdulbois: Good catch, reverted changes

deployment/terraform/gcp/public-sentries/main.tf
andkononykhin: >  X86_64 is no longer valid, but instead amd64 or arm

shouldn't we migrate to `amd64` instead of arch part removal ?

.github/workflows/release.yml
```diff
@@ -172,25 +176,22 @@ jobs:
...
           make build
           build/${{ env.BIN_NAME }} version
-
-          tar czf build/${{ env.BIN_NAME }}${{ matrix.tgz_name_suffix }} -C build ${{ env.BIN_NAME }}
         shell: bash
 
-      - name: build application and cosmovisor binary | ubuntu-22.04
-        if: ${{ matrix.os == 'ubuntu-22.04' && (!matrix.bin || !matrix.tgz || !matrix.cosmovisor) }}
+      - name: build application and cosmovisor binary | ubuntu
+        if: ${{ startsWith(matrix.os, 'ubuntu-') && (!matrix.bin || !matrix.tgz || !matrix.cosmovisor) }}
```
andkononykhin: @Abdulbois for case when `tgz` is not presented but `bin` (`dcld`) and/or `cosmovisor` are already available in the release makes sense to not rebuild the existing artifacts but rather download them so the archive will include exactly the same binaries as ones uploaded separately

old logic actually doesn't download as well

So, please add a ToDo or create a follow-up issue to cover that separately.

CC @ashcherbakov

deployment/scripts/utils/test_endpoints
```diff
@@ -35,7 +35,7 @@ Usage: $0 [persistent-endpoints-file]
 Tests connections to ip:port pairs from the persistent endpoints file.
 If the file is not specified './persistent_endpoints.txt' is used.
 
-The tool uses netcat. It tries to install it if not available!
+The tool uses netcat (netcat-openbsd). It tries to install it if not available!
```
AzizMukhtorjonov:
```suggestion
The tool uses netcat-openbsd. It tries to install it if not available!
```

Why in brackets?

README-DEV.md
```diff
@@ -1,7 +1,7 @@
 # Development Guideline
 
 Please note, that the only officially supported platform now is Linux.
-It's recommended to develop and deploy the App on Ubuntu 18.04 or Ubuntu 20.04.
+It's recommended to develop and deploy the App on Ubuntu 24.04 (20.4 and 22.04 versions are also supported).
```
AzizMukhtorjonov:
```suggestion
It's recommended to develop and deploy the App on Ubuntu 24.04. It also works on other versions of Ubuntu(20.04,22.04)
```

Saying 20.04 is supported is not fully true

docs/upgrading-ubuntu-on-running-nodes.md
```diff
@@ -0,0 +1,285 @@
...
+    5.2 Verify that the `dcld` node is running:
+    ```bash
+    dcld status
+    ```
+    
+6. **Second Release Upgrade (22.04 to 24.04)**: To upgrade to the final `24.04` release, repeat steps 2 through 5.
+   - You can skip step 1 (Snapshot).
+   - You can skip step 4.4 about opening new firewall rule for port `1022`.
+
+7. **Post-Upgrade Cleanup**: Remove the firewall rule for port `1022` from your Cloud Provider's network settings.
```
AzizMukhtorjonov, line 285: Is the process to close the port is the exact opposite of opening it? e.g. Go to the same place and only disable instead of enable?

Abdulbois: Some kind of. I mean, it should be just removed from the same place

.github/workflows/release.yml
Abdulbois: Added

deployment/scripts/utils/test_endpoints
Abdulbois: Updated

README-DEV.md
Abdulbois: Updated

AzizMukhtorjonov (approved):

deployment/terraform/gcp/public-sentries/main.tf
Abdulbois: >Because of GCP image namings for Ubuntu 24.04, X86_64 is no longer valid, but instead amd64 or arm arch are used. That's why architecture was removed

Sorry, I was wrong. So `gcloud compute images list --project=ubuntu-os-cloud --filter='architecture=X86_64 AND family=ubuntu-2404-lts'` works and returns `ubuntu-2404-noble-amd64` image. Reverted `arch` var

andkononykhin: thanks

andkononykhin (approved):

PR #716, opened 2026-04-10
#598 After removing the model version, it still appears when querying all versions of the model

### LINKED ISSUE #598: Deleting a Model-Version doesn't fully remove it from the Model
Deleting a Model-Version created under a Model doesn't correctly delete it because it can still be queried in the "versions" supported by that Model. This results in an error when trying to delete a Model after deleting one of its Model-Versions.

1. Add a Model
2. Add two Model-Versions -> add [0, 1]
3. Query all-model-versions of that model -> [0, 1]
4. Delete a Model-Version -> delete [1]
5. Query all-model-versions of that model -> [0, 1]
6. Delete model (It will still have 1 model-version ok, and another "ghost") -> "failed to execute message; message index: 0: No model version associated with vid=12288, pid=70 and softwareVersion=1 exist on the ledger: model version does not exist"
7. To delete the model, I'll need to add back that "ghost" model version  ->  add [1] again, it can be dummy as long as the Software Version matches the original that was deleted before.

Abdulbois:
- Add migration functions to remove ghost ModelVersions records
- Add new upgrade integration tests (from 1.5.2  to 1.6.0-0.dev.2)
- Cover with integration test flow 

Closes #598


ashcherbakov (approved):

integration_tests/prepare-dcld-versions.sh
```diff
@@ -24,7 +24,7 @@ log() {
 
 mkdir -p $TMP_DCLD_BINS_DIR
 
-for version in 0.12.0 0.12.1 1.2.2 1.4.3 1.4.4 1.5.1 1.5.2; do
+for version in 0.12.0 0.12.1 1.2.2 1.4.3 1.4.4 1.5.1 1.5.2 1.6.0-0.dev.2; do
```
AzizMukhtorjonov: Should it also be 1.6.0 before actual release?

Abdulbois: Yes. Left `TODO` about that

PR #720, opened 2026-04-17
fix(x509): handle hex.DecodeString error in FormatOID() instead of discarding

srpatcha:
## Summary

Malformed hex strings in certificate subject fields could previously produce **corrupt decoded output** because the error from `hex.DecodeString` was discarded with `decoded, _ := ...`. In a compliance ledger this is a security concern: an attacker-controlled or accidentally truncated certificate could yield bogus normalized OIDs in indexed fields.

## Fix

After this fix, malformed entries are skipped and the original (unmodified) subject value remains in the output, so consumers see the raw OID rather than a silently-corrupted normalized one.

```go
- decoded, _ := hex.DecodeString(value)
+ decoded, err := hex.DecodeString(value)
+ if err != nil {
+     continue
+ }
  subjectValues[index] = fmt.Sprintf("%s=0x%s", newKey, decoded)
```

## Tests

Adds two test functions to `x/pki/x509/x509_test.go` exercising the bug:

- **`Test_FormatOID_HexDecodeError`** — 5 cases: non-hex chars, odd-length hex, embedded whitespace, mixed valid/invalid OIDs, and the positive-path regression guard
- **`Test_FormatOID_NoCorruptOutput`** — explicit assertion that `vid=0x` followed by empty/null bytes cannot appear in output

Both tests **fail with the pre-fix code** and **pass with the fix applied**.

## Scope

Single concern, single commit, single author email.

### DIFF
--- a/x/pki/x509/x509.go
+++ b/x/pki/x509/x509.go
@@ -153,7 +153,10 @@ func FormatOID(header, oldKey, newKey string) string {
 			// get value from header
 			value = value[len(value)-hexStringIntegerLength:]
 
-			decoded, _ := hex.DecodeString(value)
+			decoded, err := hex.DecodeString(value)
+			if err != nil {
+				continue
+			}
 
 			subjectValues[index] = fmt.Sprintf("%s=0x%s", newKey, decoded)
 		}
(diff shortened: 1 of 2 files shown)


ashcherbakov (commented):
- It looks like the `x/audit` module shouldn't be part of this PR, should it? Can you please remove it keeping only the x509 validation fix.
- It would also be great to add some tests in https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/x/pki/x509/x509_test.go reproducing and fixing the issue.

Issue #703, opened 2025-12-18
Conformance with Matter spec


ashcherbakov:
**Acceptance Criteria:** 
1. Check  that all modules and data schemes in DCL (DeviceModel, Vendorinfo, Compliance, PKI) match the spec https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/service_device_management/DistributedComplianceLedger.adoc  - Names
  - Set of fields
  - Types
  - Validation rules
  - Other business logic
2. Fix found issues (spec or code). Think about backward compatibility while fixing.
3. Update DCL documentation with mapping of terms (in spec and in code) if there is still a difference that cannot be easily fixed without lots of breaking changes. 
4. Update DCL documentation with clarification with DCL specifics which is not mentioned in the spec.

**Examples for Items 3 an 4:**
1) what the difference is between certified-model (compliance/certified-models/<vid>/<pid>/<sw>) and compliance-info (compliance/compliance-info/<vid>/<pid>/<sw>), where each of those tables appear in the webui and how those map to the DeviceSoftwareCompliance schema as defined in the spec
2) Model -> DeviceModel in the spec and ModelVersion -> DeviceSoftwareVersionModel in the spec

PR #722, opened 2026-04-22
#703 Enable accounts with VendorAdmin role to add or modify Model and ModelVersion records

### LINKED ISSUE #719: Allow a Vendor-Admin account to write to the Model / Model-Version Schema
Discussions have identified that missing Model and Model-Version data in the DCL is largely due to members not being aware of this requirement after certification. One way to address this is to allow a Vendor-Admin account (single account, managed by Alliance staff) to populate this data on behalf of members, using only the information already submitted in the certification application. Members would still retain full control to update their records later through their approved Vendor account tied to their Vendor ID. This update would require prior review and approval in the DCL_PPG and Security Design TT.

Abdulbois:
- Enable VendorAdmin accounts add/update/delete Model and ModelVersion records
- Cover with tests
- Set Golang to exact 1.20.14 version to resolve issue(MERGE OVERLAP) in Gocovmerge CI job

Closes #719


x/model/keeper/check_model_rights.go
```diff
@@ -8,7 +8,11 @@ import (
 )
 
 func checkModelRights(ctx sdk.Context, k Keeper, signer sdk.AccAddress, vid int32, pid int32, message string) error {
-	// sender must have Vendor role to add new model
+	// a sender with a VendorAdmin role can add or modify any model
```
AzizMukhtorjonov:
```suggestion
	// sender with VendorAdmin role can add or modify any model
```

x/model/keeper/check_model_rights.go
```diff
@@ -8,7 +8,11 @@ import (
...
+	if k.dclauthKeeper.HasRole(ctx, signer, dclauthtypes.VendorAdmin) {
+		return nil
+	}
+	// the sender must have a Vendor role to add or modify a model
```
AzizMukhtorjonov:
```suggestion
	// sender must have Vendor role to add or modify a model
```

x/model/handler_test.go
```diff
@@ -178,6 +178,169 @@ func TestHandler_AddModel(t *testing.T) {
...
+	msgDeleteModel := NewMsgDeleteModel(vendorAdmin)
+	msgDeleteModel.Vid = msgCreateModel.Vid
+	msgDeleteModel.Pid = msgCreateModel.Pid
+	_, err = setup.Handler(setup.Ctx, msgDeleteModel)
+	require.NoError(t, err)
+
+	// query model
+	_, err = queryModel(setup, msgCreateModel.Vid, msgCreateModel.Pid)
+	require.Error(t, err)
+	require.Equal(t, codes.NotFound, status.Code(err))
```
AzizMukhtorjonov, line 319: Why do we expect error here?

AzizMukhtorjonov (changes requested):

x/model/handler_test.go
Abdulbois: It is expected behaviour, see `x/model/keeper/grpc_query_model.go:42`

x/model/keeper/check_model_rights.go
Abdulbois: Updated

x/model/keeper/check_model_rights.go
Abdulbois: Updated

AzizMukhtorjonov (approved):

x/model/handler_test.go
```diff
@@ -178,6 +178,169 @@ func TestHandler_AddModel(t *testing.T) {
 	require.Equal(t, msgCreateModel.CommissioningFallbackUrl, receivedModel.CommissioningFallbackUrl)
 }
```
ashcherbakov, line 180: It would also be great to add the tests when 
- Model/ModelVersion is created by VendorAdmin, and updated/deleted by Vendor
- Model/ModelVersion is created by Vendor, and updated/deleted by VendorAdmin

Abdulbois: Added

ashcherbakov (approved):

Issue #727, opened 2026-04-28
Ota fields creation and update fix


ashcherbakov:
According to the Spec OtaFileSize/OtaChecksum/OtaChecksumType are not mutable, and OtaURL is.
We have the following issue now:
1) During ModelVersion creation - no OTA fields set (no URL, no checksum, etc.)
2) During Update, URL was set, but checksum/type/filesize not.

We should support only two options for OTA fields change during update:
1) If OTA fields already set during Create: OTA URL can be present in Update, all other fields - cannot.
2) If OTA fields were not set during Create: if OTA URL is present, all other OTA fields must also be present (as in Create).

PR #728, opened 2026-04-29
#726 #727 Update creation and modification of OtaURL and its associtated fields

### LINKED ISSUE #726: otaCheckSum min value
1) we should have a validation for `otaCheckSum` from the bottom. It's not in the spec explicitly, but implicitly it's there. 
The length depends on the OtaChecksumType, and since we know what types we support (1, 7, 8, 10, 11, 12] from https://www.iana.org/assignments/named-information/named-information.xhtml#hash-alg), we can determine what min length to expect (or at least we can set a common min length for all types).
2) There is a bug in the spec. If  OtaChecksumType=12 (sha3-512), then 64 bytes is the length of the content, and base64 will have greater length. But our constraint for OtaChecksum is max 64.... Need to update the Spec and the code.

Abdulbois:
- Apply rules described in #727 while creating/updating OtaURL and its associated fields
- Set min and max length constraints for OtaChecksum
- Improve validation of OtaChecksum's value against a base64 encoded string  
- Cover with tests

Closes #726 #727


proto/zigbeealliance/distributedcomplianceledger/model/tx.proto
```diff
@@ -128,8 +128,9 @@ message MsgUpdateModelVersion {
   uint32 maxApplicableSoftwareVersion = 8 [(gogoproto.moretags) = "validate:\"omitempty,gtecsfield=MinApplicableSoftwareVersion\""];
   string releaseNotesUrl = 9 [(gogoproto.moretags) = "validate:\"omitempty,url,startswith=https:,max=256\""];
   uint64 otaFileSize = 10;
-  string otaChecksum = 11 [(gogoproto.moretags) = "validate:\"max=64\""];
+  string otaChecksum = 11 [(gogoproto.moretags) = "validate:\"max=88\""];
   uint32 schemaVersion = 12 [(gogoproto.moretags) = "validate:\"eq=0\""];
+  int32 otaChecksumType = 13 [(gogoproto.moretags) = "validate:\"gte=0,lte=65535\""];
```
AzizMukhtorjonov: Not a change request but for a history: https://www.iana.org/assignments/named-information/named-information.xhtml#hash-alg has values 0 to 63 but with protos requirements and inner checks on otaChecksumType values it is okay to keep range here and have list inside

x/model/keeper/msg_server_model_version.go
```diff
@@ -157,6 +152,29 @@ func (k msgServer) UpdateModelVersion(goCtx context.Context, msg *types.MsgUpdat
 	return &types.MsgUpdateModelVersionResponse{}, nil
 }
 
+func updateOtaFields(modelVersion *types.ModelVersion, msg *types.MsgUpdateModelVersion) error {
+	isNewOtaURL := modelVersion.OtaUrl == ""
+	if isNewOtaURL {
+		hasAllOtaMetadata := msg.OtaChecksum != "" && msg.OtaChecksumType != 0 && msg.OtaFileSize != 0
```
AzizMukhtorjonov: Revert checks and naming. It now has negative checks and then negative check on result of negative checks.

let's rename field to `missingAnyOtaMetadata` and have checks using OR rather than AND. This way we have errors if any of the fields are missing

x/model/keeper/msg_server_model_version.go
```diff
@@ -157,6 +152,29 @@ func (k msgServer) UpdateModelVersion(goCtx context.Context, msg *types.MsgUpdat
...
+		if !hasAllOtaMetadata {
+			return types.NewErrOtaMissingInformation()
+		}
+
+		modelVersion.OtaChecksum = msg.OtaChecksum
+		modelVersion.OtaChecksumType = msg.OtaChecksumType
+		modelVersion.OtaFileSize = msg.OtaFileSize
+	} else {
+		hasAnyOtaMetadata := msg.OtaChecksum != "" || msg.OtaChecksumType != 0 || msg.OtaFileSize != 0
```
AzizMukhtorjonov, line 167: Just a note: if every field to update OTA is provided - it should not silently update only URL but throw an error to show that other fields will NOT be updated

x/model/types/messages_model_version.go
```diff
@@ -228,14 +237,48 @@ func (msg *MsgDeleteModelVersion) ValidateBasic() error {
...
+			return err
+		}
+
+		_, err = base64.StdEncoding.DecodeString(ota.Checksum)
+		if err != nil {
+			return NewErrOtaChecksumIsNotBase64Encoded(ota.Checksum)
+		}
+
+		return nil
+	}
```
AzizMukhtorjonov, line 269: What about file size?

x/model/types/messages_model_version.go
```diff
@@ -228,14 +237,48 @@ func (msg *MsgDeleteModelVersion) ValidateBasic() error {
...
+
+	if ota.FileSize != 0 || ota.Checksum != "" || ota.ChecksumType != 0 {
+		return NewErrorOtaURLNotProvidedButOtherOtaFieldsProvided()
+	}
+
+	return nil
+}
```
AzizMukhtorjonov, line 276: Let's revert this check.


If ota.URL == "" & (ota.FileSize != 0 || ota.Checksum != "" || ota.ChecksumType != 0) {
    return NewErrorOtaURLNotProvidedButOtherOtaFieldsProvided()
}

		err := validateOtaChecksumType(ota.ChecksumType)
		if err != nil {
			return err
		}

		_, err = base64.StdEncoding.DecodeString(ota.Checksum)
		if err != nil {
			return NewErrOtaChecksumIsNotBase64Encoded(ota.Checksum)
		}

		return nil

x/model/types/messages_model_version_test.go
```diff
@@ -435,10 +438,12 @@ func TestMsgCreateModelVersion_ValidateBasic(t *testing.T) {
 			}(validMsgCreateModelVersion()),
 		},
 		{
-			name: "OtaFileSize == 0 when OtaUrl is omitted",
+			name: " when OtaUrl is omitted then OtaFileSize, OtaChecksumType, and  OtaChecksum must be omitted too",
```
AzizMukhtorjonov:
```suggestion
			name: "When OtaUrl is omitted then OtaFileSize, OtaChecksumType, and  OtaChecksum must be omitted too",
```

AzizMukhtorjonov (changes requested):

x/model/types/messages_model_version.go
Abdulbois: line `err := validateOtaChecksumType(ota.ChecksumType)` will fail for the case when URL is empty and ChecksumType is 0

x/model/types/messages_model_version_test.go
Abdulbois: Updated

x/model/types/messages_model_version.go
Abdulbois: For creation flow, FileSize validations should be handled within the `required_with` validation of gogoproto tags
For the update flow, within the newly added `updateOtaFields` method

x/model/keeper/msg_server_model_version.go
Abdulbois: Updated

ashcherbakov (approved):

AzizMukhtorjonov (approved):

x/model/types/messages_model_version.go
```diff
@@ -228,14 +238,58 @@ func (msg *MsgDeleteModelVersion) ValidateBasic() error {
...
+		err := validateOtaChecksumType(ota.ChecksumType)
+		if err != nil {
+			return err
+		}
+
+		if len(ota.Checksum) < 44 {
+			return errors.Wrapf(validator.ErrFieldMinLengthNotReached, "min length for OtaChecksum(base64 encoded) is 44, got %v", len(ota.Checksum))
+		}
+
+		if !base64Regex.MatchString(ota.Checksum) {
```
ashcherbakov: Why do we need this line if we have the line `base64.StdEncoding.DecodeString` below?

Abdulbois: Removed, it was redundant

ashcherbakov (approved):

PR #729, opened 2026-05-02
#724 Enable URL liveness check for release builds

### LINKED ISSUE #724: URL Reachability Check as part of CLI
Extend CLI checks when creating or updating Model/ModelVersion to verify URL reachability for the following fields: `supportUrl`, `productUrl`, `userManualUrl`, `releaseNotesUrl`, `otaUrl`, `commissioningCustomFlowUrl`.

The check can be done after the general `validate basic` static validation (in case no errors found there), see [link](https://github.com/zigbee-alliance/distributed-compliance-ledger/blob/master/x/model/client/cli/tx_model.go#L103).

Abdulbois:
-  Introduces a dev build tag to setup dev env dependent flags
- Add URL liveness checks in CLI tx command for prod release 

Closes #724


utils/validator/validations.go
```diff
@@ -49,27 +53,78 @@ func requiredIfBit0Set(fl validator.FieldLevel) bool {
 }
 
 func isValidHttpOrHttpsUrl(fl validator.FieldLevel) bool { //nolint:stylecheck
```
ashcherbakov, line 51: Reachability checks cannot be implemented here, because it's part of `ValidateBasic` function which is part of the static validation, see https://csa-matter.slack.com/archives/C025PQ6RMRA/p1777291064854469.
Please implement reachability checks in separate dedicated functions which should be called only within CLI directly, as described in the original ticket.

ashcherbakov (changes requested):

utils/validator/validations.go
Abdulbois: Update, but it is called(should be called) before the broadcasting txn to the DCLD nodes(before `tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), msg)` call)

utils/cli/url_liveness.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+const (
+	livenessCheckTimeout = 10 * time.Second
+)
+
+var allowed4XXStatusCodes = []int{
+	http.StatusUnauthorized,
+	http.StatusForbidden,
+	http.StatusUnavailableForLegalReasons,
```
ashcherbakov, line 19: Do we want to include also `405 Method Not Allowed` — server rejects HEAD specifically?
Another option can be to treat any status beside 404 (and 5xx?) as reachable.

x/model/client/cli/tx_model_version.go
```diff
@@ -59,6 +61,9 @@ func CmdCreateModelVersion() *cobra.Command {
 				specificationVersion,
 			)
 
+			if releaseNotesURL != "" && !cli.IsLiveURL(releaseNotesURL) {
```
ashcherbakov: What about `otaURL`?

x/model/client/cli/tx_model_version.go
```diff
@@ -166,6 +171,10 @@ func CmdUpdateModelVersion() *cobra.Command {
 				schemaVersion,
 			)
 
+			if releaseNotesURL != "" && !cli.IsLiveURL(releaseNotesURL) {
```
ashcherbakov: What about `otaURL`?

x/model/client/cli/tx_model_version.go
Abdulbois: Updated

x/model/client/cli/tx_model_version.go
Abdulbois: Updated

utils/cli/url_liveness.go
Abdulbois: Added 405. For the second option, as I know, when the server is down, it should return a 5xx code even requested page does not exist

.github/workflows/verify.yml
```diff
@@ -140,7 +141,8 @@ jobs:
         run: integration_tests/run-all.sh all cover
         env:
           GH_TOKEN: ${{ secrets.GH_TOKEN }}
-          GOCOVER_MODE: set  # Ensure integration tests use the same mode as unit tests
+          GOCOVER_MODE: set  # Ensure integration tests use the same mode as unit tests\
```
AzizMukhtorjonov:
```suggestion
          GOCOVER_MODE: set  # Ensure integration tests use the same mode as unit tests
```

utils/cli/url_liveness.go
```diff
@@ -0,0 +1,87 @@
...
+	http.StatusMethodNotAllowed,
+}
+var httpClient = &http.Client{Timeout: livenessCheckTimeout}
+
+func IsLiveURL(u string) bool {
+	if config.DisableURLLivenessCheck {
```
AzizMukhtorjonov: Let's rename it to `URLLivenessCheckEnabled`.
1. Mark it as a boolean flag because `Disable` looks more like function name than boolean flag.
2. Change negative to positive. It is easier to read `URLLivenessCheckEnabled` thus `!URLLivenessCheckEnabled` than `URLLivenessCheckDisabled`.

utils/cli/url_liveness.go
AzizMukhtorjonov: Yes, if the server is down - you only get server codes, not resource.

utils/cli/url_liveness.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return false
+}
+
+// FirstUnreachableURL checks the liveness of the given URLs concurrently and
+// returns the first one that is not reachable.
+// Empty strings are skipped.
+//
+// Returns an empty string if all non-empty URLs are reachable.
+func FirstUnreachableURL(urls ...string) string {
```
AzizMukhtorjonov: As far as I understand - reason of having this function is to throw error if ANY of given urls are not reachable.

It would be more UX friendly if we can give info on ALL unreachable urls in one check.

Worst case for both scenarios - timeout.
Best case for FirstUnreachableURL - when any of calls gives error.
Best case for CheckURLsForReachability - when all calls are completed.


@ashcherbakov what do you think?

Makefile
```diff
@@ -110,7 +114,8 @@ go.sum: go.mod
 	GO111MODULE=on go mod verify
 
 test:
-	go test -v $(PACKAGES)
+	go test -tags=dev -v $(PACKAGES)
+	go test -v utils/cli/url_liveness.go  utils/cli/url_liveness_test.go
```
AzizMukhtorjonov: Can we move these files into variable just like `PACKAGES`?

AzizMukhtorjonov (changes requested):

utils/cli/url_liveness.go
ashcherbakov: I think it makes sense to provide a list of all unreachable URLs at the same time for better UX.

Makefile
Abdulbois: No, because `url_liveness_test.go` should be executed without `dev` tag

utils/cli/url_liveness.go
Abdulbois: Updated

utils/cli/url_liveness.go
Abdulbois: Renamed to URLLivenessCheckEnabled

.github/workflows/verify.yml
Abdulbois: Updated

Makefile
AzizMukhtorjonov: I meant another variable.

```
PACKAGES = $(shell go list ./... | grep -v '/integration_tests')
PACKAGES_TO_RUN_WITHOUT_DEV_TAG=...
```

and then
```
	go test -tags=dev -v $(PACKAGES)
	go test -v $(PACKAGES_TO_RUN_WITHOUT_DEV_TAG)
	```

Abdulbois: I think the current way is easier. For your case, we should read the contents of all test files and filter out files where `//go: build !dev` text exists.

AzizMukhtorjonov: It is possible to just list the ones we need too like in my previous comment. This way, it would be possible to extend the list in future if we need to without extending content of the makefunction

AzizMukhtorjonov (approved):
LGTM

Makefile
Abdulbois: Updated to enable executing all test files where the `dev` tag is disabled

ashcherbakov (approved):

PR #713, opened 2026-03-05
#703 Conformance with Matter spec -  Apply new/missed constraints

Abdulbois:
- Restrict the number of active `Trustee` accounts to 12
- Reduce constraints of `Reason` field to 20KB
- Update validations and possible variants/values of`CertificationRoute`, `Transport`, `FamilyId`, `SupportedClusters`, `ProgramType` and `ParentChild` fields of the `ComplianceInfo` entity
- Add new `SpecificationVersion` into `ComplianceInfo` entity and make it as deprecated in `Model` entity
- Deperecate `CertificationIDOfSoftwareComponent`, `CompliantPlatformUsed`, `CompliantPlatformVersion` and `OSVersion` fields of `ComplianceInfo` entity
- Deprecate `SpecificationVersion` feild of `ModelVersion` entity
- Update related docs
- Make Docker and GitHub files to use the strict Golang version to resolve the issue with the coverage job

PR #730, opened 2026-05-15
#703 Bump Compliance Info schema version

Abdulbois:
- Bump ComplianceInfo to schema version 1
- Introduce a shared schema-versioning mechanism

Should be review after #713


docs/transactions/compliance.md
```diff
@@ -6,7 +6,7 @@
 
 **Status: Implemented**
 
-Attests compliance of the Model Version to the ZB or Matter standard.
+Attests compliance of a Model Version to the Matter, Zigbee or Aliro certification.
```
AzizMukhtorjonov, line 9: Shouldn't it be `the` article? It is related to ModelVersion schema and it is known schema

Abdulbois: I think the `a` article is used to mention any Model Version instance(real record), not actual schema

PR #731, opened 2026-05-18
#537 Fix issue while decoding `next_key` of pagination url (ts-client)

### LINKED ISSUE #537: Avoid using the reserved characters from RFC3986 in fields that are intended to form part of a URL (e.g. 'next_key'), or provide them already escaped.
We encountered an error during pagination due to special characters ('+' and '=') in the request URL. 

` https://on.dcl.csa-iot.org/dcl/model/models?pagination.key=AAARaC8AAAP+Lw==`

These characters cause issues because they are reserved in RFC 3986.  To prevent future errors, we need to modify DCL. It should either provide fields that can be used as URL parts without RFC 3986 characters or ensure that these characters are already properly escaped.

Abdulbois:
- Enable proper encoding of pagination URL params(keys) in ts-client
- Enable checking appield patch in CI and Makefile

Closes #537


AzizMukhtorjonov (commented):
Please provide more description for PR

AzizMukhtorjonov (approved):
lgtm

Makefile
```diff
@@ -124,9 +124,20 @@ license-check:
 clean:
 	rm -rf $(OUTPUT_DIR)
 
+# Regenerate the TypeScript client and apply the RFC 3986 query-encoding patch.
+# The patch is required so base64 pagination keys (+, /, =) survive transit;
+# without it the cosmos-sdk gRPC-gateway decodes '+' as space and rejects next_key.
+ts-client-gen:
```
ashcherbakov, line 132: Should we update Dev docs with the instructions, so that this patch is not lost during the re-generation?

Abdulbois: Added a new section about ts-client regeneration

ashcherbakov: What about sections https://github.com/Abdulbois/distributed-compliance-ledger/blob/891031738afe378ea9b8e34ec901f63f7255976f/README-DEV.md#how-to-add-a-new-module-or-transaction and https://github.com/Abdulbois/distributed-compliance-ledger/blob/891031738afe378ea9b8e34ec901f63f7255976f/README-DEV.md#how-to-make-changes-in-data-model-for-existing-modules? Should we mention a need to apply the script there?

Abdulbois: Updated

ashcherbakov (approved):

PR #740, opened 2026-06-03
Improve OS upgrade doc

Abdulbois:
- Improve the doc due to findings while upgrading SN and VN to Ubuntu 24.04

### DIFF
--- a/docs/upgrading-ubuntu-on-running-nodes.md
+++ b/docs/upgrading-ubuntu-on-running-nodes.md
@@ -6,6 +6,14 @@ This document outlines the procedure for upgrading an active Distributed Complia
 
 > **Provider scope:** The release-upgrade procedure itself (steps 3–6) is provider-agnostic. The provider-specific actions are **snapshot** (step 1), **SSH connection** (step 2), and **opening port 1022** (step 4.4). Expand the section for your cloud provider in each step.
 
+> **Multi-node order (Sentry Node + Validator Node):** Upgrade **sentries first (one at a time), then the validator last**.
+>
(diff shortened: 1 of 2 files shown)


docs/upgrading-ubuntu-on-running-nodes.md
```diff
@@ -6,6 +6,14 @@ This document outlines the procedure for upgrading an active Distributed Complia
 
 > **Provider scope:** The release-upgrade procedure itself (steps 3–6) is provider-agnostic. The provider-specific actions are **snapshot** (step 1), **SSH connection** (step 2), and **opening port 1022** (step 4.4). Expand the section for your cloud provider in each step.
 
+> **Multi-node order (Sentry Node + Validator Node):** Upgrade **sentries first (one at a time), then the validator last**.
+>
+> Why it's safe: this repo's ansible deploys the validator with `pex: false` and `persistent_peers` set only to its private sentries (see `deployment/ansible/roles/configure/vars/validator.yml` and `tasks/validator.yml`). Taking one sentry down leaves the validator connected through the others, so signing continues.
```
ashcherbakov: > Taking one sentry down leaves the validator connected through the others, so signing continues.

But ``pex: false` and `persistent_peers` set only to its private sentries` - doesn't it mean that Validator has connection to the Sentry only (they are in a private net)? So there is no "others"?

Abdulbois: Clarified statement. Please check

ashcherbakov (approved):

AzizMukhtorjonov (approved):
lgtm

